PR #3, opened 2021-07-03
Revert "Revert "Removed file_id in Span, compact file sources""

jonathandturner:
Reverts jonathandturner/engine-q#2


jonathandturner:
Seeing some phantom benchmark issues. This didn't actually regress as much as it seemed, and in fact seems a tiny bit better.

PR #6, opened 2021-08-23
Implement ranges

kubouch:
This PR starts with a primitive integer range parsing but I plan adding the full ranges from Nushell (and maybe extend with a step). I'm creating this PR early to check if I'm doing it right.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2882,5 +2882,24 @@ mod tests {
                 })
             ));
         }
+
+        #[test]
+        fn parse_subexpression_range() {
+            let parser_state = ParserState::new();
+            let mut working_set = ParserWorkingSet::new(&parser_state);
+
+            let (block, err) = working_set.parse_source(b"(3 - 3)..<(8 + 2)", true);
+
+            assert!(err.is_none());
+            assert!(block.len() == 1);
+            assert!(matches!(
+                block[0],
+                Statement::Expression(Expression {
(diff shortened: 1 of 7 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -1030,6 +1038,80 @@ impl<'a> ParserWorkingSet<'a> {
...
+
+        if bounds.len() != 2 {
+            return (
+                garbage(span),
+                Some(ParseError::Expected("range".into(), span)),
+            );
+        }
+
+        let span_lhs = Span::new(span.start, span.start + dotdot_pos[0]);
+        let lhs = match self.parse_number(bounds[0], span_lhs) {
```
jntrnr, line 1084: I think for this you want to recursively parse the expressions on either side. For example, it could be `(2 + 3)..(5 + 9)` and both sides would have a subexpression

jntrnr:
I think once you have the recursion thing fixed, you've got it. The other stuff I looked at seemed right

jntrnr:
@kubouch - out of curiosity, what's left on the list?

kubouch:
@jntrnr There was a panic when I type `(0).."a"`, I want to investigate that. Then, I'll resolve merge conflicts and mark for merge.

Other features I want to add:
* negative numbers
* floats
* stepping
but I'll add them as separate PRs because I hate merge conflicts. I'll create a tracking issue for these to keep it organized.

kubouch:
I made a new PR

PR #8, opened 2021-08-31
Comments with a newline dont get together

elferherrera:
This PR adds tests for lite parser to check that comments are separated in different statements

### DIFF
--- a/crates/nu-parser/tests/test_lite_parser.rs
+++ b/crates/nu-parser/tests/test_lite_parser.rs
@@ -16,20 +16,25 @@ fn lite_parse_helper(input: &[u8]) -> Result<LiteBlock, ParseError> {
 
 #[test]
 fn comment_before() -> Result<(), ParseError> {
+    // Code:
+    // # this is a comment
+    // def foo bar
     let input = b"# this is a comment\ndef foo bar";
 
     let lite_block = lite_parse_helper(input)?;
 
-    assert_eq!(lite_block.block.len(), 1);
+    assert_eq!(lite_block.block.len(), 2);
     assert_eq!(lite_block.block[0].commands.len(), 1);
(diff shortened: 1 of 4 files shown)


crates/nu-parser/tests/test_lite_parser.rs
```diff
@@ -44,39 +49,63 @@ fn comment_beside() -> Result<(), ParseError> {
 
 #[test]
 fn comments_stack() -> Result<(), ParseError> {
+    // Code:
+    // # this is a comment
+    // # another comment
+    // # def foo bar
```
jntrnr, line 55: I wonder if for cases like this we should treat the comments as a simple block of comments

elferherrera: that is a good point. Should those comments be part of the next command or should they be a separate statement?

jntrnr: We might have to play with it a bit to know if that's what we want, but I was thinking they should be part of the next command.

elferherrera: Sounds like a good plan. So, we could start saying that all comments above a command should be attached to it. 
What about this case?
```
# comment 1
# comment 2
let a = ( ls |
where name =~ some |   # another comment
each { |file| rm file.name } # final comment
```
or this
```
# comment 1
# comment 2
let a = ( ls 
| where name =~ some    # another comment
| each { |file| rm file.name } # final comment
```

jntrnr: @elferherrera I think for the second case they'd attach to the things that come before. For the first case, maybe both attach to the second one?

In practice, you'll only really see the ones that are doc comments I think, and they'll be on the definition parts rather than the expressions.

elferherrera:
@jntrnr can you have a look at the changes I did the lite-parser?

PR #12, opened 2021-09-04
Similar name check for signature

elferherrera:
Check if a signature is created with similar name. If a similar name is found, then the function panics


jntrnr:
Nice!

PR #16, opened 2021-09-04
Implement positive integer ranges

kubouch:
Implements the following:
* `0..3` -> [0, 1, 2, 3]
* `0..<3` -> [0, 1, 2]
* `98..` -> [98, 99, 100]
* `98..<` -> [98, 99]
* subexpressions and variables as ranges bounds 
  * `let a = 2; $a..(4 + $a)` -> [2, 3, 4, 5, 6]
* syntax highlighting
* one step towards stepping
  * `0..2..10` detects the step in the middle but still parses it as `0..10`

Known bugs:
* `..10` doesn't work
* syntax highlighting for stepped ranges doesn't work for the middle part (might be resolved once stepping is fully implemented)
* works only for integers >= 0 and the range must be ascending

### DIFF
--- a/crates/nu-cli/src/syntax_highlight.rs
+++ b/crates/nu-cli/src/syntax_highlight.rs
@@ -57,6 +57,10 @@ impl Highlighter for NuHighlighter {
                 FlatShape::Float => {
                     output.push((Style::new().fg(nu_ansi_term::Color::Green), next_token))
                 }
+                FlatShape::Range => output.push((
+                    Style::new().fg(nu_ansi_term::Color::LightPurple),
+                    next_token,
+                )),
                 FlatShape::Bool => {
                     output.push((Style::new().fg(nu_ansi_term::Color::LightCyan), next_token))
                 }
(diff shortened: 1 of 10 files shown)


kubouch:
I plan to implement the missing features in future PRs to avoid merge conflicts since engine-q tends to move quite fast sometimes (and actually write down what the missing features are lol).

jntrnr:
Sounds good. Code looks okay, just some CI issues to fix I think.

kubouch:
@jntrnr With the last commit, I made somewhat controversial change to allow parsing for negative numbers and left-unbounded ranges. Do you think it's ok like that?

jntrnr:
Nice!

Starting to brainstorm how we can make this easy to work with from the commands since we've got a lot of different types of things that you can iterate through.

kubouch:
@jntrnr Currently, the `..10`, and previously `-1`, errors with a cryptic "unsupported operation" error. I haven't looked at the `parse_call()` code closely, but perhaps we could try putting a `parse_math_expression()` inside if the call was unsuccessful.

PR #34, opened 2021-09-08
Move value into its own folder in nu-protocol and add some comments

moonrise-tk:


jntrnr:
Yeah makes sense, especially if Value keeps growing as we add support for more things

PR #40, opened 2021-09-10
Parse errors for def, let and alias

elferherrera:
Parsing no longer panics with def, let and alias

image

### DIFF
--- a/crates/nu-cli/src/errors.rs
+++ b/crates/nu-cli/src/errors.rs
@@ -209,7 +209,7 @@ pub fn report_parsing_error(
                 Diagnostic::error()
                     .with_message("Unknown state")
                     .with_labels(vec![
-                        Label::primary(diag_file_id, diag_range).with_message(format!("{}", name))
+                        Label::primary(diag_file_id, diag_range).with_message(name.to_string())
                     ])
             }
             ParseError::NonUtf8(span) => {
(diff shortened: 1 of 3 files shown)


elferherrera:
@jntrnr I changed the error and now it looks like this
image

I had to do it like this because in this [function](https://github.com/nushell/engine-q/blob/main/crates/nu-parser/src/parser.rs#L2542) any error that is produced by the let, alias and def functions is ignored and passed to the `else` clause.

I tried a function like this one

```rust
pub fn parse_statement(
    working_set: &mut StateWorkingSet,
    spans: &[Span],
) -> (Statement, Option<ParseError>) {
    let name = working_set.get_span_contents(spans[0]);

    match name {
        b"def" => parse_def(working_set, spans),
        b"let" => parse_let(working_set, spans),
        b"alias" => parse_alias(working_set, spans),
        _ => {
            let (expr, err) = parse_expression(working_set, spans);
            (Statement::Pipeline(Pipeline::from_vec(vec![expr])), err)
        }
    }
}
```
that the syntax highlighter can use to warn you are you type those functions. As seen in the original image. However, doing it like this, the error we get from `parse_def`, `parse_let` or `parse_alias` is that those declarations are not found, specially when running the parse tests, since they are done without any declaration.

What do you think about it?

jntrnr:
We can keep iterating on the exact errors. I can land this as an improvement on what we have.

I'd probably change the "internal error" to something else since internal error means an error that occurred because nushell itself broke. You might have seen an ICE when doing Rust - which is Internal Compiler Error (same general idea)

For these, these are just normal errors we could tell the user what parts are missing.

elferherrera:
I can change that as well. I left the internal error since I saw it in the parser

PR #42, opened 2021-09-11
added source command

moonrise-tk:
Creates a source command to execute shell files. Not sure if I did this right. Some notes:

* There is some code from the nushell version that I don't understand. I left it as comments at line 53 in `source.rs`. I don't know if there is currently any equivalent in engine-q
* I added nu-parser as a dependency for nu-command, in order to parse the source file. I don't know if this is unwanted.
* It doesn't work lol. Gives me the following error: 
```sh
error: Type mismatch
  ┌─ line_1:1:8
  │
1 │ source "./example.sh"
  │        ^^^^^^^^^^^^^^ expected FilePath, found String
```
Probably need to mess w/ parser somewhere? Not sure.

### DIFF
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -8,3 +8,4 @@ edition = "2018"
 [dependencies]
 nu-protocol = { path = "../nu-protocol" }
 nu-engine = { path = "../nu-engine" }
+nu-parser = {path = "../nu-parser"}
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -5,7 +5,7 @@ use nu_protocol::{
     Signature, SyntaxShape,
 };
 
-use crate::{Alias, Benchmark, BuildString, Def, Do, Each, For, If, Length, Let, LetEnv};
+use crate::{Alias, Benchmark, BuildString, Def, Do, Each, For, If, Length, Let, LetEnv, Source};
 
 pub fn create_default_context() -> Rc<RefCell<EngineState>> {
     let engine_state = Rc::new(RefCell::new(EngineState::new()));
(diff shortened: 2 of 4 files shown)


kubouch:
@moonrise-tk The actual sourcing happens at parse time, see here: https://github.com/nushell/nushell/blob/0fa0c25fb3e2758282c151a378ab82d64a025445/crates/nu-parser/src/parse.rs#L1871 .

The way Nushell works is that it first parses _everything_, then evaluates _everything_. Parsing at eval time is a no-no but you'd need to ask JT about the theory behind it.

Honestly, I'm not sure what all the logic in the `source` command is for. I used to think it is just to have an entry in the `help commands` message, but maybe there's more to it.

jntrnr:
@kubouch - it looks like some of the current nushell logic for `source` is for properly tagging spans we `source`. In engine-q, this should get a lot easier as e-q should do that for you. You just tell it what file to parse and it will update the spans properly.

In theory, in e-q you can use the same `parse` function during parse time (at the bottom of the parse.rs) and tell it to not use scoping (passing false for the `scoped` parameter). This will tell e-q to put the definitions it finds in the current scope.

This is all *in theory*... until we try it we won't know for sure :sweat_smile:

jntrnr:
Oh yeah, and the Command part of it gives the Signature, which we use for `help` and syntax highlighting.

jntrnr:
@moonrise-tk - any luck with this one?

moonrise-tk:
@jntrnr was on vacation for the past week, so didn't get much done. I just pushed the work I had before: it still doesn't compile, but hopefully its closer to what you would do?

moonrise-tk:
Comments from discord: Needs to run the file (this happens at eval-time)

PR #43, opened 2021-09-11
Add stepping support & reversing to ranges

kubouch:
Follows the following syntax: `<start>..<next-value>..<end>`. In case some of the values is missing, it gets automatically inferred.

Also small refactor to the range iterator.

Examples (these should go into test suite once we have eval tests set up):
* `2..4..8` -> `[2, 4, 6, 8]`
* `2..4..<6` -> `[2, 4, 6]`
* `2..-2..-8` -> `[2, -2, -6]`
* `..2..6` -> parse error but should be equivalent to `0..2..6`
* `2..4..` -> equals to `2..4..100`
* `2..-4..` -> equals to `2..-4..-100`
* `2..6` -> `[2, 3, 4, 5, 6]`
* `2..-2` -> `[2, 1, 0, -1, -2]`
* `2..-2..-2` -> `[2, -2]`

Integer values only for now.

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -85,13 +85,19 @@ pub fn eval_expression(
             val: *f,
             span: expr.span,
         }),
-        Expr::Range(from, to, operator) => {
-            // TODO: Embed the min/max into Range and set max to be the true max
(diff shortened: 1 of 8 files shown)


jntrnr:
nice!

PR #45, opened 2021-09-12
Allow parsing left-unbounded range (..10)

kubouch:
This allows parsing `..10` and similar as a left-unbounded range.

Some points:
* Valid path should not be parseable as a valid range because the only case you can have a path separator in a range is this: `0..(4 / 2)..10` which involves spaces. Paths should not have spaces around the separator (I hope?).
* `..` is not allowed to be parsed as a range. Range always needs at least one bound, otherwise it could be easily confused with parent directory.
* With this change, *external* commands with names that could be a valid range (e.g. `..10`) are not allowed since they would parse as a range.
* *Internal* commands named `..10` etc. are allowed.


jntrnr:
looks good

PR #47, opened 2021-09-13
more block param and build string tests in concert with lists

stormasm:
added some more tests around the idea of block params in concert with list iteration

added some more build-string tests using lists


jntrnr:
thanks!

PR #51, opened 2021-09-19
External with input

elferherrera:
Externals now accept input from previous statements in the pipeline. This will be handy if we want to "feed" the external command via stdin

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -98,7 +98,7 @@ fn eval_external(
         })
         .collect();
 
-    command.run(&context, &call, input)
+    command.run(context, &call, input)
 }
 
 pub fn eval_expression(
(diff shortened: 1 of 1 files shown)


crates/nu-engine/src/eval.rs
```diff
@@ -68,6 +68,40 @@ fn eval_call(context: &EvaluationContext, call: &Call, input: Value) -> Result<V
...
+    input: Value,
+) -> Result<Value, ShellError> {
+    let engine_state = context.engine_state.borrow();
+
+    let decl_id = engine_state
+        .find_decl("run_external".as_bytes())
+        .ok_or_else(|| ShellError::ExternalNotSupported(*name))?;
+
+    let command = engine_state.get_decl(decl_id);
+    let new_context = context.enter_scope();
```
jntrnr, line 84: I don't think you need to enter the scope here, since the external doesn't have it's own block

PR #52, opened 2021-09-20
replace codespan-reporting with miette 3.0

zkat:
Before:

image

After:

Screen Shot 2021-09-20 at 11 29 42


zkat:
@jntrnr test updated :)

jntrnr:
Thank you! Looking forward to playing with this :)

PR #55, opened 2021-09-21
use miette's new panic hook

zkat:
This also changes the style of depending on miette to target specific commits. Not only will this make things more predictable, but also you won't have to do some weird `cargo update --workspace -p miette` any time you want the latest miette.


jntrnr:
let's do it - we can improve where this lights up in follow-up changes

PR #61, opened 2021-09-23
Externals with redirection

elferherrera:
PR introduces the option to redirect the output from an external command to a ValueStream.
It also adds a simple version of the command lines to check the ValueStream


jntrnr:
oh nice!

PR #62, opened 2021-09-24
Better print out for stream output

elferherrera:
Corrected bug with channel. Using `sync_channel` was blocking some external commands, like `rg`.

Things that still need correction:
- When using `cat` (probably only windows) the prompt gets mangled and it can only be seen as bytes. It goes back to normal after typing something and pressing enter
image

- Python commands are not sent to stdout. If we send `python -c "print('hello')"` the output is missed. However we can enter a python shell without problem

- The use of `^` for marking and external command. Now `git` cannot be used because there is a git command in engine-q

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -234,9 +234,9 @@ dependencies = [
 
 [[package]]
 name = "instant"
-version = "0.1.10"
+version = "0.1.11"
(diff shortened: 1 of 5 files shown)


jntrnr:
In Linux, passing args to Python seems okay:

```
〉python3 -c "print('hello')" | each { $"($it) is good" }
[hello
 is good]
```

PR #69, opened 2021-09-26
Primitive module implementation

kubouch:
Introduces `module` and `use` keywords to define a module and import custom commands from it.

Current limitations:
* Currently, only custom commands are supported for export
* Also, all defs within a module are exported, there is no selective export yet, neither are the definitions hidden behind namespace
* `module` also closes over its parent scope which we probably want to disallow
  * This could be handled by the same mechanic as deactivate: When entering a module scope, deactivate everything
* No module from file yet
* No deactivate yet
* No nested modules yet
* Modules are implemented as Blocks which expose what definitions are to be exported. However, we do not use the underlying Block anywhere at all, so we might change Module to be just a list of exportables.
  * If we make `use` to also evaluate the module, we would probably need the block

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -7,7 +7,7 @@ use nu_protocol::{
 
 use crate::{
     where_::Where, Alias, Benchmark, BuildString, Def, Do, Each, External, For, Git, GitCheckout,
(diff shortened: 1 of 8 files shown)


jntrnr:
looks good!

PR #74, opened 2021-09-28
Modules: export & hide

kubouch:
This PR introduces two keywords:
* `export` for controlling what to export from a module
* `hide` used to hide the definition in the current scope

Maybe a  screenshot describes it the best:
modules2

Needs some polish and tests. There might be edge cases.


crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -124,9 +129,23 @@ impl EngineState {
...
 
     pub fn find_decl(&self, name: &[u8]) -> Option<DeclId> {
+        let mut hiding_strength = 0;
+        // println!("state: starting finding {}", String::from_utf8_lossy(&name));
+
         for scope in self.scope.iter().rev() {
+            // println!("hiding map: {:?}", scope.hiding);
+            // check if we're hiding the declin this scope
+            if let Some(strength) = scope.hiding.get(name) {
+                hiding_strength += strength;
```
jntrnr, line 139: Does the hiding strength here mean how many levels of the scope (that contain this name) that we'd need to go through to find the first un-hidden one?

jntrnr: Oh, I see, so this lets you also do:

```
hide foo
hide foo
```

And know you want to hide two versions of foo

kubouch: @jntrnr  Yes, that's the point. There is a bug currently, that you cannot unhide a def (by `def foo [] {}`) -- you should be able to do that.

Also it panics on this:
```
> def foo [] { "foo" }

> do {
  hide foo
  def foo [] { "bar" }
}
```

Yeah, needs to be tweaked...

jntrnr:
Looks good

PR #84, opened 2021-10-02
Prompt with env variable

elferherrera:
Introduces the ability to configure the prompt using the `PROMPT_COMMAND` env variable

image


jntrnr:
nice

PR #85, opened 2021-10-02
add serde derive to Cargo.toml so nu-protocol compiles standalone

stormasm:
We want to be able to compile nu-protocol (standalone) by itself and run the nu-protocol tests.

By adding the derive feature to serde this works...


jntrnr:
nice catch

PR #88, opened 2021-10-04
port the mv command

xiuxiu62:
A port of the move command with the `move_file` and `move_item` methods pulled from nushell/nu-engine's filesystem module.  I'm not sure if there are plans to move similar functions into the engine crate at some point, but this is my first pr on the project and I didn't want to be presumptuous.


crates/nu-command/src/filesystem/mv.rs
```diff
@@ -0,0 +1,141 @@
...
+                "can only move multiple sources if destination is a directory".to_string(),
+            ));
+        }
+
+        let some_if_source_is_destination = sources
+            .iter()
+            .find(|f| matches!(f, Ok(f) if destination.starts_with(f)));
+        if destination.exists() && destination.is_dir() && sources.len() == 1 {
+            if let Some(Ok(filename)) = some_if_source_is_destination {
+                return Err(ShellError::InternalError(format!(
```
jntrnr, line 71: Nice!

My one suggestion is that instead of using ShellError::InternalError, can you add cases to ShellError for these different errors?

If you can't see how to do it, let me know and I can make a few suggestions.

xiuxiu62: Awesome I was wondering about that.  I'll poke around implementations in other commands and get something together.  Will definitely reach out if I need help, thanks.

jntrnr:
Looks good - I think the last thing is to run this through `cargo clippy`. Once the warnings are fixed it should be able to go green.

xiuxiu62:
Okay i'm hoping i did that correctly.  This is only my second time contributing to an open source rust project and appreciate the help lol.

jntrnr:
Looks good!

PR #96, opened 2021-10-05
WIP: output `ls` as a grid vs table

fdncred:
This is a work in process. I won't be able to work on it for a while (maybe months) so I thought I'd throw it up here for someone else to take a stab at it if it looks interesting to them.

image

This integrates `term_grid` which doesn't look supported any longer. I integrated all the code in grid.rs because I had to change one of the files due to a bug.

still to do:
- [ ] hook up other Value::* inputs in the match statement, if needed
- [ ] somehow add the ability to still use ` | where type == File` functionality
- [ ] maybe allow the user to specifiy the column delimiter - i just have it hard coded as `|` but in bash there is no delimiter. maybe some people would want the normal bash look.
- [ ] maybe allow the user to specify more of the `GridOptions` settings
- [ ] maybe eventually allow a config point where `ls` is always output to a grid vs a table? not sure - or maybe aliasing it will be enough
- [ ] ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 history.txt
 /target
+/.vscode
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -524,7 +524,9 @@ dependencies = [
  "nu-protocol",
  "nu-table",
  "sysinfo",
+ "terminal_size",
  "thiserror",
+ "unicode-width",
 ]
 
 [[package]]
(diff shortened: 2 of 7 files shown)


jntrnr:
@fdncred you okay with me stealing this for reedline too?

fdncred:
Please take it and do it right. I was hoping someone would jump in.

PR #98, opened 2021-10-05
port `cp` to fs commands

xiuxiu62:
This also includes a few fs utils for handling nested directory mapping and creation.  I added a few std::convert::From impls for ShellErrors as well, which are resolving to internal errors right now, but can probably be swapped out for more descriptive errors.


xiuxiu62:
Yeah I read a a thread about a similar problem.  It looks like there are some situations where it can be a pretty non-descriptive error message, but we should be able to just drop the collect altogether here.

jntrnr:
Thanks!

PR #107, opened 2021-10-08
Add `in` and `not-in` operators support

arthur-targaryen:
## What it does
- Port `in` and `not-in` operators
- Implement `PartialOrd` for `Value`
- Add `Range::contains` method

## To-Do

- [x] `in` operator
  - [x] string in string
  - [x] number in range
  - [x] anything in list of anything
  - [x] string in record
  - [x] anything in stream of anything
  - [ ] *other?*
- [x] `not-in` operator
  - [x] string in string
  - [x] number in range
  - [x] anything in list of anything
  - [x] string in record
  - [x] anything in stream of anything
  - [ ] *other?*


fdncred:
you probably want to add value in table. [this is how](https://github.com/nushell/nushell/blob/b6728efcd4330b3b1d7dcabc410b259d13ef4bdf/crates/nu-engine/src/evaluate/operator.rs#L92) it's done in nushell

PR #114, opened 2021-10-10
add `rm`

xiuxiu62:
This pr adds a few things, mainly a port of the `rm` command.  I additionally introduced the bind_command macro for easily binding boxed commands to the engine instance and put stubs in for the `open` and `save` commands.  Please please please look over this carefully as it's a pretty critical command and I haven't been sleeping well.  I haven't been able to fully ditch value streams in the helper method's return, as flat_map seems to want an implementation that i haven't been able to figure out yet.  I would love some help smoothing the edges, but hopefully the base looks solid enough.


xiuxiu62:
I'm not sure why this is failing.  I moved the closure into a variable just to grab the type class in my text editor, but moved it back in the last commit.  I'll take a deeper look; sorry if I'm missing something,

jntrnr:
@xiuxiu62 looks like you need to run `cargo fmt` on the code. The CI will check that all the code uses the standard Rust formatting.

crates/nu-command/src/filesystem/open.rs
```diff
@@ -0,0 +1,53 @@
+use nu_protocol::ast::Call;
```
jntrnr, line 1: For commands we aren't ready to implement, probably better to leave them out for now so we don't accidentally think they're done

jntrnr: I'm reworking the `.map` now, so might have an easier way to do it in the future. Probably worth landing that part as is and then revisiting as more helpers are added.

xiuxiu62: oh nice, yeah the issue now is that it will return a stream of errors as a value to the the client rather throwing any errors.  Is that alright in the meantime?

jntrnr:
Looks good!

PR #115, opened 2021-10-10
Hiding rehaul

kubouch:
Fixes the "hiding" system so that you can call def/use/hide repeatedly and get the expected result.

See commit messages for details.


jntrnr:
Looks good!

PR #230, opened 2021-10-12
add size command

xiuxiu62:


jntrnr:
if you're curious, I did a couple small touch-ups to bring up the unit tests too

PR #235, opened 2021-10-14
cp, mv, and rm commands need to support -i flag

GabrielBG0:
Implements interactive flag to copy, move and remove file system commands
Related to [#3896](https://github.com/nushell/nushell/issues/3162#issuecomment-943692540) on nushell

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,6 +155,21 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


jntrnr:
Looks good. You'll need to run this through `cargo clippy` and fix the warnings (as well as `cargo fmt` if you haven't yet)

PR #239, opened 2021-10-17
added custom completion

rsteube:
basic idea is to somehow enable registration of custom completers which are invoked and passed the current command line up to cursor position to then return completions as json (or similar)

- uses reedline with `Suggestion` struct for descriptions during completion
- uses dialoguer with added paging to `fuzzy_select` (still some issues with line removal)
- still hardcoded call instead of configuration which command to invoke for completion (should be sth. like https://github.com/rsteube/nushell/blob/0.38.0-a986de8ad_carapace2/crates/nu-completion/src/custom.rs)
- still quite a mess

this currently works with [carapace-bin](https://github.com/rsteube/carapace-bin) after adding the binary to PATH

related: https://github.com/nushell/reedline/pull/167
related: https://github.com/mitsuhiko/dialoguer/compare/master...rsteube:fuzzy-paging

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "dialoguer"
+version = "0.9.0"
+source = "git+https://github.com/rsteube/dialoguer?branch=fuzzy-paging#92b59e2ca8aed50ed8ebc7e094aa8c7e7f2b6d99"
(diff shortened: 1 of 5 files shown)


fdncred:
this was cool to play with, thanks for allowing us to play and test with it. however, it's quite out of date now so it's probably time to close it. thanks!

PR #243, opened 2021-10-19
Loading modules from files

### LINKED ISSUE #258: Not possible to call multi-word commands
Steps to reproduce:

1. `def "foo bar" [] { "foo" }` 
2. `foo bar` --> error: command not found

This is possible in Nushell.

kubouch:
### New import pattern syntax
```
> module spam { def foo [] { "foo" } }

> use spam foo

> foo
```

### Loading modules from files

```
> bash -c "echo 'export def foo [] { \"foo\" }' > spam.nu"

> use spam.nu

> spam foo
foo

> use spam.nu foo

> foo
foo

> hide foo

> hide spam *
```

### Makes it possible to call multi-word commands

Fixes #258.

### Fixes wrong file contents offsets in the working set

See changes in `engine_state.rs` and a commit message in [7112664b](https://github.com/nushell/engine-q/pull/243/commits/7112664b3fb9caceb72bb6d302d33a5637ee0910)

### Left for later:

Currently, the module loaded from file closes over existing scope, which means that local definitions will be visible. We probably do not want that but fixing it would require some thinking -- e.g., we should make some default "overlay" that would be visible but hide the user-defined stuff. So I'll leave it for a future PR.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1698,7 +1698,7 @@ pub fn parse_import_pattern(
     let source = working_set.get_span_contents(span);
     let mut error = None;
 
-    let (tokens, err) = lex(source, span.start, &[], &[b'.']);
+    let (tokens, err) = lex(source, span.start, &[], &[b':']);
     error = error.or(err);
 
     if tokens.is_empty() {
@@ -1713,7 +1713,7 @@ pub fn parse_import_pattern(
 
     let head = working_set.get_span_contents(tokens[0].span).to_vec();
 
-    if let Some(tail) = tokens.get(2) {
+    if let Some(tail) = tokens.get(3) {
         // FIXME: expand this to handle deeper imports once we support module imports
         let tail_span = tail.span;
         let tail = working_set.get_span_contents(tail.span);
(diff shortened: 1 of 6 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -613,112 +642,90 @@ pub fn parse_call(
...
+        name_spans.push(*word_span);
 
-    if let Some(mut decl_id) = working_set.find_decl(name) {
-        let mut name = name.to_vec();
-        while pos < spans.len() {
-            // look to see if it's a subcommand
-            let mut new_name = name.to_vec();
-            new_name.push(b' ');
-            new_name.extend(working_set.get_span_contents(spans[pos]));
+        let name = working_set.get_span_contents(span(&name_spans));
```
jntrnr, line 663: So far, I've been treating `math  sum` the same as `math sum`. With this change, the user will always need to type the latter I think?

Probably good for things to be uniform, but wanted to mention it in case we want the old support.

jntrnr:
Yeah, I think this is cool to merge subcommands/modules like this. Just had the one question about the change in how we handle the names. If we're okay with that, I think this is good to land.

kubouch:
Just hit a bug with multi-word commands (#269) but it's not specifically about files so I would fix that in another PR.

jntrnr:
Cool, let's do that. This one looks good, then

PR #245, opened 2021-10-22
update readme with the issue 242 where people can sign up for commands to port

stormasm:
Some extra marketing material on our engine-q Readme file for folks who stop by and want to know where to sign up at !


jntrnr:
if you pull in the latest main for this and the other PR, in theory they should go green. I think the hang up was a clippy issue

stormasm:
@jntrnr thanks I sent them both through again and they went green, thank you for fixing the clippy issues !

Issue #242, opened 2021-10-19
Porting commands to engine-q list


stormasm:
Updated on January 14, 2022

### Instead of commenting on this issue, please go to #735 

Engine-q will become Nushell's new parsing and evaluation engine in the future. As part of this, we'll port the existing Nushell commands into Engine-q. Much of this porting can be done mechanically, as while the two systems differ, they are similar enough that many commands can be ported following the steps below.

You can sign up to help out porting a command by commenting on this issue.

Our preference when sending us a Pull Request (PR) is to do one command at a time.  It makes it a lot easier for us to review the code as well as making the merge conflicts easier to figure out...

**Note:** Instructions on how to port these commands from Nushell to Engine-q are **below the list**.

|name|sub|category|completed|notes|
|-|-|-|-|-|
|alias||core|done||
|all?|||done|@arthur-targaryen|
|ansi|||done|@matthewauld|
|ansi|gradient||done|@matthewauld|
|ansi|strip||done|@matthewauld|
|any?|||done|@arthur-targaryen|
|append||filters|done|@stormasm|
|autoenv||||@jntrnr|
|autoenv|trust|||@jntrnr|
|autoenv|untrust|||@jntrnr|
|autoview|||done|@jntrnr - mostly built into `table` now|
|benchmark||system|done||
|build-string||strings|done||
|cal|||done|@onthebridgetonowhere|
|cd||filesystem|almost done|missing `cd -`|
|char|||done|@matthewauld|
|chart|||||
|clear|||done|@Scorpil|
|collect|||done||
|compact|||done|@matthewauld|
|cp||filesystem|done||
|dataframe|||done|@elferherrera|
|date|||done|@onthebridgetonowhere|
|date|format||done|@onthebridgetonowhere|
|date|humanize||done|@onthebridgetonowhere|
|date|list-timezone||done|@onthebridgetonowhere|
|date|now||done|@onthebridgetonowhere|
|date|to-table||done|@onthebridgetonowhere|
|date|to-timezone||done|@onthebridgetonowhere|
|debug|||done||
|def||core|done||
|default||||@matthewauld|
|describe|||done||
|do||core|done||
|drop||filters|done|@stormasm|
|drop|column|filters|done|@stormasm|
|drop|nth|filters|done|@stormasm|
|du||||@Scorpil|
|each||filters|done||
|each|group|filters|||
|each|window|filters|||
|echo||core|done|@aslynatilla|
|empty?|||done|@onthebridgetonowhere|
[...]

PR #268, opened 2021-10-31
Port date commands to enqine-q

onthebridgetonowhere:
I ported the date commands to engine-q. One thing I am not sure about is the `to-table` command. The result is a "column" table and not a "row" table. Is my understanding correct that 1-row tables are automatically pivoted to vertical tables?

Looking forward to feedback :-)

Issue reference: https://github.com/nushell/engine-q/issues/242


jntrnr:
@onthebridgetonowhere - in engine-q, there's finally a difference between a record and a one-row table. If you make a Record and put it in a Stream or List, you'll get the one row table.

jntrnr:
@onthebridgetonowhere - just reviewed it, well done! this looks great.

Last step is that you'll need to run through `cargo clippy --all` and fix the warnings you see. With that I think we're good.

onthebridgetonowhere:
@jntrnr thanks for the review and suggestions. Fixed the one-row table issue as well as all clippy issues.

jntrnr:
Heh, this is reminding me we should use nicer test assertions :D  Looks like the difference is between Record and List of a single Record on the failing test.

onthebridgetonowhere:
Fixed the failing test case.

jntrnr:
looks good!

PR #270, opened 2021-11-01
Plugins for engine q

elferherrera:
This PR introduces the missing plugin concept for engine-q. As considered in nushell, a plugin is a binary that extends engine-q abilities. It gives us the ability to "attach" components to the engine without having to compile them into engine-q's code. 

A quick example of a plugin for engine-q written in rust would look like this

```
use nu_plugin::{plugin::PluginError, serve_plugin, Plugin};
use nu_protocol::{ast::Call, Signature, SyntaxShape, Value};

struct Test;

impl Plugin for Test {
    fn signature(&self) -> Signature {
        Signature::build("test-plugin")
            .required("first", SyntaxShape::String, "first argument")
            .required("second", SyntaxShape::Int, "second argument")
    }

    fn run(&self, _call: &Call, input: &Value) -> Result<Value, PluginError> {
        Ok(input.clone())
    }
}

fn main() {
    serve_plugin(&mut Test)
}
```
To register the plugin to engine-q, it has to be compiled (cargo build --release) and the location of the resulting binary is used with the register command like this:
```
register <path-to-binary>
```
This will add a new declaration into the engine, which means that your plugin is ready to be used

Things to do:
- [ ] Expand the types of arguments that can be used as signatures (need access to Stack) 
- [ ] Expand the list of values that can be passed to the plugin 
- [ ] Store registered commands in file so they dont have to be registered all the time when engine-q starts


jntrnr:
I think this looks good. I'm not familiar with capnproto. A couple questions:

- Does this support bidirectional interaction with the engine? Like can it ask the engine to evaluate a block and get the value back?
- Should we port one of the existing plugins so we have something we can run and test? maybe `inc`?

elferherrera:
I think we can have inter process communication. We could define some rpc functions and use them to communicate between them.

Regarding migrating one plugin, what structure do you suggest for the plugins? Should we make them part of engine-q repo or should they be in another project in nushell folder?

jntrnr:
@elferherrera I was just thinking of having one plugin ported for testing purposes. If you don't think we need it and that the other testing is enough, I think we're good

elferherrera:
I think I will port it. Im just wondering how should we include it in the repo. Should it be in the plugin crate as an example?

jntrnr:
@elferherrera for now, how about just another crate in the crates dir like we do in nushell?

jntrnr:
let's do this!

PR #271, opened 2021-11-01
Porting echo command

aslynatilla:
Implementation of command `echo`, assigned in #242.
I also ran `cargo clippy --all`, so that I was sure I didn't introduce unnecessary warnings.

I'll be waiting for feedback, before going on with `format`.

### DIFF
--- a/crates/nu-command/src/core_commands/echo.rs
+++ b/crates/nu-command/src/core_commands/echo.rs
@@ -1,7 +1,6 @@
 use nu_protocol::ast::Call;
 use nu_protocol::engine::{Command, EngineState, Stack};
-use nu_protocol::{PipelineData, ShellError, Signature, SyntaxShape};
-//TODO: add Example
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/core_commands/echo.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    fn run(
+        &self,
+        _engine_state: &EngineState,
+        _stack: &mut Stack,
+        _call: &Call,
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        Ok(input)
```
jntrnr, line 28: Apologies, looks like I misspoke a little. The arguments from the `rest` argument are what should be turned into the output.

aslynatilla: No problem!
I'll deal with it in some minutes! Thanks for the feedback!

jntrnr:
thanks!

PR #277, opened 2021-11-02
Port first command

onthebridgetonowhere:
Ported the first command. 

@jntrnr - a couple of questions :-) 
* the original Nushell code had some matching on Primitive::Binary, which I did not fully understand.  Is that needed? My implementation is very simple, similar to the `last` command implementation
* Do we have use cases where we want the first (x) elements of a string? E.g., "test" | first --> t or "test" | first 2 --> te. At the moment, if the input is a string we get back the whole string. 

Thanks!


jntrnr:
This looks good. We can think about binary and string support after we spend some time with it.

I can also put a note on the issue that binary support can be added later as well.

fdncred:
we definitely want binary support in `first` and binary support in general. think about this use case. i want to see the magic number in a binary file. i only want to see the first few bytes, not the entire file.

i also think having `$string | first` is a cool addition.

PR #278, opened 2021-11-02
Port into string command

onthebridgetonowhere:
I ported most of `into string` command. We do not support the older column paths at the moment. I'd be happy to work more on this if I understand what's the output of `into string` when called on a table. It makes the specified column's values into strings?

### DIFF
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -22,6 +22,10 @@ unicode-segmentation = "1.8.0"
 glob = "0.3.0"
 thiserror = "1.0.29"
 sysinfo = "0.20.4"
+bigdecimal = { package = "bigdecimal-rs", version = "0.2.1", features = ["serde"] }
+num-bigint = { version="0.3.1", features=["serde"] }
+num-format = { version="0.4.0", features=["with-num-bigint"] }
+num-traits = "0.2.14"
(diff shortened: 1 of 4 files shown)


crates/nu-command/Cargo.toml
```diff
@@ -22,6 +22,10 @@ unicode-segmentation = "1.8.0"
 glob = "0.3.0"
 thiserror = "1.0.29"
 sysinfo = "0.20.4"
+bigdecimal = { package = "bigdecimal-rs", version = "0.2.1", features = ["serde"] }
+num-bigint = { version="0.3.1", features=["serde"] }
+num-format = { version="0.4.0", features=["with-num-bigint"] }
+num-traits = "0.2.14"
```
jntrnr, line 28: Oops, I should have mentioned in the porting guide - engine-q doesn't use bigint or bigdecimal like nushell. We've moved to i64/f64. For porting, if you use the i64/f64 methods, you should be good.

jntrnr:
I did something similar with column paths - I think we could leave them out for now and then revisit the `into` group later on and see how we want to add support in a future PR.

crates/nu-command/Cargo.toml
onthebridgetonowhere: Fixed it :)

jntrnr:
looks good!

PR #282, opened 2021-11-03
math: floor, ceil, median and mode

luccasmmg:
Port for the `math floor`, `math ceil`, `math median`, and `math mode` commands.

The first three are very straightforward.

```
〉[1.5 2.4 -3.1] | math floor
╭───┬────╮
│ 0 │ 1  │
│ 1 │ 2  │
│ 2 │ -4 │
╰───┴────╯
```
```
〉[1.5 2.4 -3.1] | math ceil
╭───┬────╮
│ 0 │ 2  │
│ 1 │ 3  │
│ 2 │ -3 │
╰───┴────╯
```
```
〉ls | math median
╭──────────┬────────────────╮
│ modified │ 41 minutes ago │
│ type     │ file           │
│ name     │ main.rs        │
│ size     │ 11.8 KB        │
╰──────────┴────────────────╯
```
```
〉[1 2 3 3] | math median
2.5
```

`math mode` on the other hand was quite tricky, the old implementation used a HashMap to store the values as keys with the frequency of them appearing being the value. The problem with that is that this requires the entry/key  to implement Eq and Hash, which in the old one(Using the primitive types) it did, but now that all the Values have different structures that is no longer possible(some don't even have a val inside). So my workaround this was create a new type that stored the bytes representation of the numbers(f64 doesn't implement Hash), and methods to convert from and convert back to the `Value` type. The main problem of this implementation is two, one is kind of ugly(in my opinion) and two because it lost the ability to work with other types(Strings for instance), i can technically make the `HashableType` more complex to acomodate this, but i was hoping someone would have a more elegant solution to propose. 
The other problem with `math mode` was its requirement of the `sort-by` command, which is not yet implemented, there are other commands left to port that have the same problem, for now i decided to just write the sorting code inside the function(is quite simple) and in the future, if i have the time i will take the job to port the `sort-by` or will wait for someone else to port it.

I would love some sugestions, i feel like my solution maybe not the best one.  

Btw, here are some examples of `math mode` use: 

```
〉[[size, height, filesize]; [100, 40, 1GB] [60, 30, 1GB]] | math mode
╭──────────┬─────────────╮
│ height   │ [30, 40]    │
│ size     │ [60, 100]   │
│ filesize │ [1000.0 MB] │
╰──────────┴─────────────╯
```

```
〉ls | math mode
╭──────┬──────────╮
│ size │ [4.1 KB] │
╰──────┴──────────╯
```

```
〉[1 1 2 2 3] | math mode
╭───┬───╮
│ 0 │ 1 │
│ 1 │ 2 │
╰───┴───╯
```

```
〉[1 1 2 2 1day] | math mode
Error: nu::shell::type_mismatch (link)

  × Type mismatch during operation.
   ╭─[entry #7:1:1]
 1 │ [1 1 2 2 1day] | math mode
   ·        ┬  ─┬─    ────┬────
   ·        │   │         ╰── type mismatch for operator
   ·        │   ╰── duration
   ·        ╰── int
   ╰────

```


crates/nu-command/src/math/mode.rs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        return Err(values);
+    }
+    //In e-q, Value doesn't implement Hash or Eq, so we have to get the values inside
+    // But f64 doesn't implement Hash, so we get the binary representation to use as
+    // key in the HashMap
+    let hashable_values: Result<Vec<HashableType>, ShellError> = values
+        .iter()
+        .map(|val| match val {
+            Value::Int { val, .. } => Ok(HashableType::new(val.to_be_bytes(), NumberTypes::Int)),
```
fdncred, line 93: assuming big endian may not be good. there are some functions in nushell to figure this out.
```rust
fn int_to_endian(n: i64) -> Vec<u8> {
    if cfg!(target_endian = "little") {
        n.to_le_bytes().to_vec()
    } else {
        n.to_be_bytes().to_vec()
    }
}

fn bigint_to_endian(n: &BigInt) -> Vec<u8> {
    if cfg!(target_endian = "little") {
        n.to_bytes_le().1
    } else {
        n.to_bytes_be().1
    }
}
```

luccasmmg: Because of the different target architectures right? That makes sense, i'm not used to having to think about bytes, thanks, gonna wait if there is a different solution that don't involve a bytes representation, and if there isn't im going to change to your sugestion.

jntrnr: We could see how hard it would be to implement Hash on Value. Would that do it?

luccasmmg: We would need to impl Hash and Eq(It would be relatively, we would probably hash and compare for equality just the values and ignore the spans) and in the case of `Value::Float` we would try to hash and compare the bytes representation, but, we would have problem in some corner cases like 0.2 + 0.1 not being equal to 0.3. i can do it if you want, just let me known where to put the code.

jntrnr: Since we're just using this for a hash, being big endian or little endian might also not matter just because the hash just needs some bits. (it'd likely matter if we tried to use it in other ways)

@fdncred - you okay with this landing if the be/le stuff is just for the hash and nothing else?

fdncred: `ne_bytes` looks good to me

jntrnr:
Okay cool, looks like we're good. Thanks @luccasmmg!

PR #284, opened 2021-11-03
Add back binary support for the first command

onthebridgetonowhere:
Add back binary support for the first command. I have not tested it as I am not sure how :-) Suggestions are welcomed.


fdncred:
Thanks! I think it would need to be piped to a version of `binaryview` like in nushell.

jntrnr:
Hehe, yeah, we'll need maybe a simple hex output at some point. This should be good for now, and we can go back and check it.

@onthebridgetonowhere - do you want to file an issue to remind us to check it after the hex viewer is in?

onthebridgetonowhere:
@jntrnr yes, I can file the issue.

PR #287, opened 2021-11-04
Port str case commands

onthebridgetonowhere:
Still need to support the column paths. I guess maybe we should figure how to do this soon :)


jntrnr:
Agreed, will see if we can add that first, so we don't have so many Todos 😅

jntrnr:
Taking a look now

jntrnr:
@onthebridgetonowhere - okay, I added some support for paths to columns: https://github.com/nushell/engine-q/pull/290

onthebridgetonowhere:
Thanks, I'll have a look!

jntrnr:
This looks great! Can you also throw a test in the sec/tests.rs file (or as an example to a command) that tests the optional cell path?

onthebridgetonowhere:
Added a `cell path` test for each `str case` command in `src/tests.rs`. Should be good to go now :-)

src/tests.rs
```diff
@@ -849,3 +849,43 @@ fn range_and_reduction() -> TestResult {
...
+    run_test(
+        r#"([[lang, gems]; [nuTest, 100]]).lang | get 0 | str snake-case"#,
+        "nu_test",
+    )
+}
+
+#[test]
+fn str_case_screaming_snake_cell_path() -> TestResult {
+    run_test(
+        r#"([[lang, gems]; [nuTest, 100]]).lang | get 0 | str screaming-snake-case"#,
```
jntrnr, line 888: Thanks for adding some tests here. Can we change some of these to take in a column path? These are using the default but not taking in an argument that gives the column to update.

onthebridgetonowhere: Oh, I see, I was not sure how to do this. Thanks for checking and clarifying it!

onthebridgetonowhere: Fixed it!

jntrnr:
looks good!

PR #288, opened 2021-11-04
Multiple commands per plugin

elferherrera:
With this PR it is possible to declare multiple signatures in one plugin. When the plugin get registered the signatures are converted into native declarations that can be used to call the specific functions in the plugin

A simplified version of the `inc` plugin was incorporated to check how a plugin can be created


jntrnr:
This looks good!

PR #290, opened 2021-11-05
Add updating cell paths

jntrnr:


crates/nu-command/src/conversions/into/binary.rs
```diff
@@ -129,27 +132,27 @@ fn float_to_endian(n: f64) -> Vec<u8> {
...
 }
 
-pub fn action(input: Value, span: Span) -> Value {
+pub fn action(input: &Value, span: Span) -> Value {
     match input {
-        Value::Binary { .. } => input,
+        Value::Binary { .. } => input.clone(),
         Value::Int { val, .. } => Value::Binary {
-            val: int_to_endian(val),
+            val: int_to_endian(*val),
```
fdncred, line 139: @jntrnr should all these be `*_ne_bytes()` too, or does that not apply here?

jntrnr: yeah, it seems the function does change based on the target, but we could probably simplify it to the `ne_bytes` version

PR #301, opened 2021-11-06
Fix "`math sum` doesn't support streams"

### LINKED ISSUE #299: `math sum` doesn't support streams
Looks like streams current assume that we're looking at a table. There are a couple other cases we should add. For example, if you have a stream of simpler values, it'd be nice to add those.

Something like:
```
[1] | each { $it } | math sum
```
cc @luccasmmg

luccasmmg:
Fixes #299  Before the math commands assumed that a stream would automatically be a table therefore something similar to a list of records, but there are other cases such as `[1] | each { $it } | math sum` when what we have is not a list of records, we run the `mf` directly into the values. 

```
〉[1 2 3] | each { $it } | math sum
6
```


jntrnr:
Thanks!

PR #303, opened 2021-11-07
Port over the reverse command from nushell

stormasm:
This adds in the ability in engine-q to reverse a table or list similar to functionality we find in nushell...


stormasm:
@jntrnr the failures are saying I should write the code like this...
```rust
        //let v: Vec<_> = input.into_iter().collect();
        let iter = input.into_iter().rev();
        Ok(iter.into_pipeline_data(engine_state.ctrlc.clone()))
```

When I do that the compiler is telling me...

```rust
let iter = input.into_iter().rev();
                             ^^^ the trait `DoubleEndedIterator` is not implemented for `PipelineIterator`
```

How would you suggest I move forward ?

jntrnr:
@stormasm - the clippy::needless-collect lint is buggy. We've reported issues on it in the past, but at least so far I don't think it's been fixed.

Go ahead and turn the lint off. We do that in a few other places too

stormasm:
@jntrnr Thanks for the tip !  That worked...

PR #317, opened 2021-11-09
Port str capitalize

onthebridgetonowhere:


jntrnr:
Looks good!

fdncred:
I wish we had a better name for this because `capitalize` is a bit ambiguous to me. i.e. 
* capitalize what? the entire string? nope, that's to-upper (which is what upcase should be named, of course). 
* capitalize each word? nope, that's title case and this doesn't do that.
* capitalize only the first letter of the first work? yup, that's what this does - now what's a better name for this? - hehe

onthebridgetonowhere:
@fdncred in principle I agree with you, sometimes it is slightly confusing. However, capitalize or capitalisation does refer to a word where its first character is upper case (or all characters uppercase). I think that's why it is important good examples; it helps being clear what the command does.

PR #319, opened 2021-11-09
Port str downcase and str contains

onthebridgetonowhere:
Two more commands ported. I was wondering if the desired output for `str contains` on tables is to really modify the column's value to `true/false`. I could see also potential use cases where we need to know which rows contain that string pattern, and then this does not work nicely. Any thoughts?


jntrnr:
looks good!

PR #326, opened 2021-11-10
Add record literal syntax

jntrnr:
This allows for a JSON-like record syntax:

```
let x = {"b": 5}
```

As well as hjson-like relaxed syntax:

```
let x = {b: 5}
```


hustcer:
@jntrnr I like this feature! Besides, can we merge two records simply like JS `Object.assign()`?

PR #331, opened 2021-11-13
Add environment variable support for modules

kubouch:
Lets you export env vars from a module and use/hide them.

(EDIT: The env var hiding shown on the screenshot was removed but is still a planned feature)

export_env


kubouch:
I think we can try to merge this and continue with the env hiding / overlay stuff in the next PR, since this one is getting stale. The ignored tests point at the relevant problems now.

For some reason, it broke the multi-word export fix...

jntrnr:
sounds good

PR #332, opened 2021-11-13
Add a config variable with engine support

jntrnr:
This adds a new builtin variable `$config`. The `$config` variable allows users to update the config by setting this to a record with known fields.

I currently have only ported over a couple settings, but we can keep adding more.

My current config.nu:

```
let $config = {
  filesize_metric: $true
  table_mode: with_love
}
```

The `$config` variable follows all the scoping of normal variables and is assumed captured by blocks.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -35,7 +35,7 @@
 - [x] ctrl-c support
 - [x] operator overflow
 - [x] Support for `$in`
-- [ ] config system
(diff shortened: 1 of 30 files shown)


jntrnr:
Going ahead and landing. We can improve this in future updates, but probably good to porting some more code over to using it.

PR #338, opened 2021-11-14
Port parse command

CBenoit:
I'm not exactly sure I did the right thing with spans and errors. Also, am I right with the list of records? This looks like a lot of duplicated information.


jntrnr:
@CBenoit - nice! 

Looks like you'll need to run `cargo clippy --all` and fix any of the warnings you see.

CBenoit:
Done! :+1:

PR #339, opened 2021-11-15
delete row.rs in nu-protocol/value which has references to RowStream

stormasm:
delete the file row.rs
this also completely eliminates 

```rust
std::{cell::RefCell, rc::Rc};
```
from the codebase which was part of a previous legacy engine-q architecture.


jntrnr:
Nice find

PR #343, opened 2021-11-16
Category option for signature

elferherrera:
Added a category method for signature to split commands by their type

### DIFF
--- a/crates/nu-command/src/conversions/into/binary.rs
+++ b/crates/nu-command/src/conversions/into/binary.rs
@@ -2,7 +2,7 @@ use nu_engine::CallExt;
 use nu_protocol::{
     ast::{Call, CellPath},
     engine::{Command, EngineState, Stack},
-    Example, PipelineData, ShellError, Signature, Span, SyntaxShape, Value,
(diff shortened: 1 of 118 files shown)


jntrnr:
nice

fdncred:
this is great! if we're going for parity with nushell, we'll have to circle back at some point and add `is_custom` for the custom commands and the signature. i'm not exactly sure what `creates_scope` indicates for our users and i notice it's only true for the `for` command. I like having the extra detail.

one last thing i noticed, `help commands` doesn't highlight the word `commands` like a sub command.

jntrnr:
We might also want to group all the flags together into their own record so we don't need quite so many columns

fdncred:
ya, agreed. this pr exposed a problem in our new table functionality, which is, it doesn't have the `...` column when it's too wide, yet.

jntrnr:
@fdncred hmm, it should have it. it's a port of `nu-table` from nushell. I wonder what's tripping it up

fdncred:
It didn't work in windows. Did the ... work for you in Linux!

PR #344, opened 2021-11-16
Separate Overlay into its own thing

kubouch:
Overlay is no longer attached to a Block.
Makes access to overlays more streamlined by removing this one indirection.
Also makes it easier to create standalone overlays without a block which might come in handy.

This is a precursor to reliable hiding of exported environment variables. The current plan is that in the Stack, the env vars would reference OverlayId they belong to.


jntrnr:
Ah okay, now I see what you mean. Will be fun to play with this

PR #346, opened 2021-11-18
Fix term width for the table

jntrnr:
Do a similar calculation for width to what we do in nushell

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -36,17 +36,36 @@
 - [x] operator overflow
 - [x] Support for `$in`
 - [x] config system
+- [x] plugins
+- [ ] external plugin signatures
+- [ ] external command signatures
 - [ ] shells
 - [ ] autoenv
-- [ ] plugins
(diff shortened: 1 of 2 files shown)


fdncred:
nice! glad you found this. let me ask another question. i think i did this btw - but e-q uses `terminal_size` instead of `term_size`, which nushell uses. I noticed `miette` uses `term_size` as well, so we already have a dependency on it, indirectly i guess. do we want to switch to `term_size` or just leave it. i'm guessing i picked `terminal_size` because it was update more recently. also because `term_size` has been archived.

jntrnr:
@fdncred yeah, we might want to PR the change to Miette to move to the more supported thing.

zkat:
@fdncred @jntrnr FYI I noticed this issue and just released `miette@3.3.0`, which uses `terminal_size` instead.

jntrnr:
@zkat - thank you!

PR #349, opened 2021-11-18
Plugins signature load

elferherrera:
The signatures for the plugins are stored in file `plugin.nu` which are used every time nushell starts. This reduces the loading time since it isnt required to call to each plugin to get its signature


jntrnr:
Nice! Let's give this a go

PR #352, opened 2021-11-19
`from xlsx` `from ods` and `from toml`

luccasmmg:
They all work pretty much as in the old engine

```
〉cat ./Cargo.toml | from toml | get features | get plugin
╭───┬───────────────────╮
│ 0 │ nu-plugin         │
│ 1 │ nu-parser/plugin  │
│ 2 │ nu-command/plugin │
╰───┴───────────────────╯
```

```
〉cat ./Teste.xlsx | from xlsx -s [Planilha1] | get Planilha1
╭───┬─────────┬─────────╮
│ # │ Column0 │ Column1 │
├───┼─────────┼─────────┤
│ 0 │ foo     │ bar     │
│ 1 │ 1       │ 2       │
╰───┴─────────┴─────────╯
```

```
〉cat ./TesteODS.ods | from ods -s [Planilha1] | get Planilha1
╭───┬─────────┬─────────┬─────────╮
│ # │ Column0 │ Column1 │ Column2 │
├───┼─────────┼─────────┼─────────┤
│ 0 │ 1       │ 2       │ 3       │
│ 1 │ yo      │ Yo 2    │ Yo 3    │
╰───┴─────────┴─────────┴─────────╯
```

A few things to consider. 
-1: I couldn't add tests examples to `from xlsx` and `from ods` because they use binary output which would be quite weird to add and to appear on the `-h` page. 
-2: In the old engine, `from xslx` had a flag `-n`. Which was there to `don't treat the first row as column names`. The thing is, this is the default behaviour anyway, so i did some diging in how the codebase looked back when the command was added(28 of november of 2019) and it turns out this flag never did anything, i don't know why it was there. Maybe in the future i'm going to add a flag which adds the behaviour of treating the first row as the column name, but for now i think i will focus on porting all the rest of the `from` commands if thats ok.


jntrnr:
looks good!

PR #358, opened 2021-11-20
port over the drop command from nushell

stormasm:
reproduce the functionality of the nushell drop command...

once this PR lands then I will start working on the....

**subcommands**

drop column
drop nth


jntrnr:
looks good

PR #362, opened 2021-11-23
Hiding of environment variables

kubouch:
This changes environment variables to be arranged in a stack-like structure so that the values can be recovered from the parent scopes. This allows us to implement deactivating overlays.

To detect at eval time whether a parser hid a definition under the same name or not, I kind of hacked the import pattern but I couldn't think of a better way that wouldn't be overly complicated.


jntrnr:
:+1: nice!

PR #365, opened 2021-11-24
Port `all?` command

arthur-targaryen:
### Main changes

- Port the `all?` command from nushell

### Other changes

- Implement `From<bool>` for `Value, see 03667f3f.
- Add `PipelineData::into_iterruptible_iter` method, see 498079c9.
- Change `IntoPipelineData` and `IntoInterruptiblePipelineData` bounds to respectively accept `Into<Value>` and `IntoIterator`, see a1fd7077.

```rust
impl<V> IntoPipelineData for V
where
   V: Into<Value>,
{
    // --snip--
}

impl<I> IntoInterruptiblePipelineData for I
where
    I: IntoIterator + Send + 'static,
    I::IntoIter: Send + 'static,
    <I::IntoIter as Iterator>::Item: Into<Value>,
{
    // --snip--
}
```

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -20,6 +20,7 @@ pub fn create_default_context() -> EngineState {
         // TODO: sort default context items categorically
         bind_command!(
             Alias,
+            All,
             Append,
             Benchmark,
             BuildString,
--- a/crates/nu-command/src/filters/all.rs
+++ b/crates/nu-command/src/filters/all.rs
@@ -1,8 +1,9 @@
 use nu_engine::eval_expression;
-use nu_protocol::ast::{Call, Expr, Expression};
-use nu_protocol::engine::{Command, EngineState, Stack};
-use nu_protocol::IntoPipelineData;
-use nu_protocol::{Category, Example, PipelineData, ShellError, Signature, SyntaxShape};
+use nu_protocol::{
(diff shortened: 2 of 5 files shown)


jntrnr:
The way I was thinking of the helpers was that they could map from a single value to a single value, or an iterator of values to an iterator of values.

For things like `any?` you can probably skip the helpers and just call `.into_iter()` on the `PipelineData` as it comes in.

```
input.into_iter().any(|x| ...)
```

You probably don't need to add any special case helpers for it.

arthur-targaryen:
> The way I was thinking of the helpers was that they could map from a single value to a single value, or an iterator of values to an iterator of values.
> 
> For things like `any?` you can probably skip the helpers and just call `.into_iter()` on the `PipelineData` as it comes in.
>
> You probably don't need to add any special case helpers for it.

All right I understand, it makes sense. I'll change this.

jntrnr:
@arthur-targaryen - taking a look to see if we can update `RowCondition` to also be a block

jntrnr:
@arthur-targaryen - I updated `RowCondition` - https://github.com/nushell/engine-q/pull/368

arthur-targaryen:
> I updated `RowCondition` - https://github.com/nushell/engine-q/pull/368

Nice, thanks I'll merge `main` into my branch and use it!

jntrnr:
looks good

PR #370, opened 2021-11-27
Plugin trait

elferherrera:
PR adds simple arguments for plugins (ints and strings) that can be passed to the plugin without calling the EngineState or the stack. 

There is a new `nu_plugin_example` that shows how to get values from the arguments for a plugin and how to call different functions within the plugin

The trait FromValue was moved from nu-engine to nu-protocol to make it available to other crates that need to convert Values to basic types


elferherrera:
Im going to redo this approach by evaluating all arguments and passing them as values to the plugin

PR #371, opened 2021-11-27
Port over the sleep command from nushell

Scorpil:
Two minor things to note:

- I didn't copy over this comment: https://github.com/nushell/nushell/blob/b8e2bdd6b11b2881b832a2b21d24cc78d7056080/crates/nu-command/src/commands/platform/sleep.rs#L47, as it seems obsolete. Correct me if I'm wrong.
- It always returns nothing directly, rather then misusing iterator trait (as there's no actual iteration taking place).


jntrnr:
Looks good!

PR #373, opened 2021-11-27
Port over the clear command from nushell

Scorpil:


elferherrera:
Thanks. Looks good

fdncred:
I've been thinking a bit about this. Since engine-q is using crossterm should we be using crossterms `clear()` function instead of shelling out to the OS to clear the screen?

elferherrera:
@fdncred we may have to find a way to pass reedline api to the command to have access to crossterm

Scorpil:
Good point @fdncred, I wasn't aware we have crossterms. I might look into that after porting other commands I signed up for.

fdncred:
> @fdncred we may have to find a way to pass reedline api to the command to have access to crossterm

ya, it would be good to be able to access reedline's crossterm instance. there are a few things i'd like to do with it without requiring another dependency. having said that, we already have a dependency on crossterm due to some of my code as well as we'll eventually need it to port andres' chart commands (can't remember if those are plugins or not).

PR #374, opened 2021-11-27
Option to replace command same name

elferherrera:
This adds the ability for commands to replace others based on the Category of the input that is passed to the pipe

image

image


elferherrera:
Moved the dataframe files to the `nu-command` crate. I am trying to consolidate all dataframe information (object definitions and commands) in a single place. Im planning to allow only dataframe commands to use dataframe objects

crates/nu-command/src/dataframe/open.rs
```diff
@@ -15,15 +15,15 @@ pub struct OpenDataFrame;
 
 impl Command for OpenDataFrame {
     fn name(&self) -> &str {
-        "open-df"
+        "open df"
```
jntrnr, line 18: this isn't going to be good for the name, as `df` is a valid filename.

jntrnr: though maybe that's less of an issue with this since? it's hard to tell

PR #379, opened 2021-11-28
Testing suite for dataframes

elferherrera:
- Implements the dataframe testing suite for dataframe commands
- Dataframe command files are moved to the nu-dataframe folder. Dataframe commands will be organized based on the CustomCommand that uses them


jntrnr:
:+1:

PR #381, opened 2021-11-29
Port over the kill command from nushell

Scorpil:
*Note*: extended `call.rs` with `get_named_arg` that returns spanned named argument for the error message. Let me know if you have a better idea for a name of that function.

Screenshot 2021-11-29 at 11 19 41


crates/nu-command/src/platform/kill.rs
```diff
@@ -0,0 +1,149 @@
...
+            let mut cmd = CommandSys::new("kill");
+            if force {
+                if let Some(Spanned {
+                    item: _,
+                    span: signal_span,
+                }) = signal
+                {
+                    return Err(ShellError::IncompatibleParameters {
+                        left_message: "force".to_string(),
+                        left_span: call.get_named_arg("force").unwrap().span,
```
jntrnr, line 87: Ahh, we don't have `unwrap()` prevention yet. I'll try to add that to the CI.

You might want to refactor to remove those. I was also wondering if there was a way to use the `.get_flag(..)` of the style of line 56 above

jntrnr: Actually, don't worry about the `unwrap()`. It looks like a bunch crept in since we didn't have the clippy flag to check for them.

Scorpil: > I was also wondering if there was a way to use the .get_flag(..) of the style of line 56 above

`get_flag` only works for arguments with values like `--signal 123`, and returns spanned value (e.g. 123). It does not work for switches (will always return None), since there is no value and no corresponding span.

The only alternative I can think of without `get_named_arg` would be just returning `ShellError::IncompatibleParametersSingle` with `call.head` span, but that's less precise.

jntrnr: Sure, using a helper is fine then.

jntrnr:
thanks!

PR #387, opened 2021-11-30
Fix busy poll with reedline

### LINKED ISSUE #386: idle engine-q uses too much cpu
I'm not sure why, but I've noticed recently that when engine-q is running and doing nothing, not even animating the prompt, it's using too much cpu. It's always around 99%.

Screen Shot 2021-11-30 at 3 40 35 AM

For comparison, here's idle nushell.
Screen Shot 2021-11-30 at 3 46 39 AM

sholderbach:
Fixes #386

Makes the changes to accept https://github.com/nushell/reedline/pull/188

Change CLI option `EQ_PROMPT_ANIMATE_MS` to binary `EQ_PROMPT_ANIMATE`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2125,7 +2125,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.2.0"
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks!

PR #389, opened 2021-11-30
Add the support of str to-int to the into int command

onthebridgetonowhere:
@fdncred, @kubouch 
this PR adds the `str to-int` support for the `into int` command. Let me know if you have any feedback.

### DIFF
--- a/crates/nu-command/src/conversions/into/int.rs
+++ b/crates/nu-command/src/conversions/into/int.rs
@@ -5,6 +5,11 @@ use nu_protocol::{
     Category, Example, PipelineData, ShellError, Signature, Span, SyntaxShape, Value,
 };
 
+struct Arguments {
+    radix: Option<Value>,
+    column_paths: Vec<CellPath>,
+}
+
 #[derive(Clone)]
 pub struct SubCommand;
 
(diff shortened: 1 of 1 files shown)


fdncred:
nice! so `into int` now does eveything it did before plus everything `str to-int` did?

crates/nu-command/src/conversions/into/int.rs
```diff
@@ -134,28 +181,120 @@ pub fn action(input: &Value, span: Span) -> Value {
...
+        let expected = Value::test_int(10);
+
+        let actual = action(&word, Span::unknown(), 10);
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn turns_binary_to_integer() {
+        let s = Value::test_string("0b101");
+        let actual = action(&s, Span::unknown(), 10);
```
jntrnr, line 281: I'm surprised this doesn't have to be a 2 instead of a 10. Do you know why it works, by chance?

onthebridgetonowhere: Good catch. When a string is matched, we check if the string starts with `0b`. If yes, then the "given" radix is ignored and 2 is used instead. Same thing for a string starting with "0x", we default to 16. You can see this in line 188 and line 216. 

I think the code can be slightly improved.

onthebridgetonowhere:
@fdncred yes. I added all the previous test cases of `str to-int` to `into int`, so it should be complete now.

jntrnr:
thanks! We can always improve things but I think this is good

PR #392, opened 2021-12-01
add optional footer to table

fdncred:
This adds the ability to have an optional footer at the end of a table. The footer is a duplicate of the header. It can be configured by placing this in your `config.nu` file.
`  footer_mode: always`

The options are:
```rust
pub enum FooterMode {
    /// Never show the footer
    Never,
    /// Always show the footer
    Always,
    /// Only show the footer if there are more than RowCount rows
    RowCount(u64),
    /// Calculate the screen height, calculate row count, if display will be bigger than screen, add the footer
    Auto,
}
```
`Auto` is not implemented yet and is just ignored.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1549,6 +1549,7 @@ version = "0.36.0"
 dependencies = [
  "ansi-cut",
  "nu-ansi-term 0.39.0",
+ "nu-protocol",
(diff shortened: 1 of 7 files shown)


onthebridgetonowhere:
This looks really nice!

PR #394, opened 2021-12-01
Port str trim

onthebridgetonowhere:
Larger PR due to many test cases. Feedback is welcomed!

### DIFF
--- a/crates/nu-command/src/strings/str_/trim/command.rs
+++ b/crates/nu-command/src/strings/str_/trim/command.rs
@@ -1,0 +1,1160 @@
+use nu_engine::CallExt;
+use nu_protocol::{
+    ast::{Call, CellPath},
+    engine::{Command, EngineState, Stack},
+    Example, PipelineData, ShellError, Signature, Span, SyntaxShape, Value,
+};
+use std::sync::Arc;
+
+#[derive(Clone)]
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/strings/str_/trim/command.rs
```diff
@@ -0,0 +1,1160 @@
...
+    call: &Call,
+    input: PipelineData,
+    trim_operation: &'static F,
+) -> Result<nu_protocol::PipelineData, nu_protocol::ShellError>
+where
+    F: Fn(&str, Option<char>, &ClosureFlags) -> String + Send + Sync + 'static,
+{
+    let head = call.head;
+    let (options, closure_flags, input) = (
+        Arc::new(Arguments {
```
jntrnr, line 129: You may be able to remove the Arc and just let the iterator move `Arguments` here

jntrnr:
I'm going to go ahead and land since I think it's good. Just saw the Arc and wondered if we needed it.

crates/nu-command/src/strings/str_/trim/command.rs
onthebridgetonowhere: Good catch, I missed that one :) I will fix it in another PR as this one was already merged. Thanks!

PR #396, opened 2021-12-02
`to url` and `to toml`

luccasmmg:
`to url` will take a table with strings in its cells and output a string formatted as a url. The behaviour is pretty much the same as in the old engine. 

```
〉[[foo bar]; ["1" "2"]] | to url
foo=1&bar=2
```
If the cells inside are not strings an error will be given, i can change this behaviour to automatically convert the values to strings, i chose not to do it just so it would have the same output as the old engine.

```
〉[[foo bar]; [1 2]] | to url
Error: nu::shell::unsupported_input (https://docs.rs/nu-protocol/0.1.0/nu-protocol/enum.ShellError.html#variant.UnsupportedInput)

  × Unsupported input
   ╭─[entry #2:1:1]
 1 │ [[foo bar]; [1 2]] | to url
   ·                      ───┬──
   ·                         ╰── Expected table with string values
   ╰────

```

`to toml` is a little bit different given the fact that the we don't have `Table` and `Row` anymore, and now we have `List` and `Record`. 

The top level can now be a "table" represented as a `List` of `Records`.
Ex: 

```
〉[[foo bar]; [1 2]] | to toml
bar = 2
foo = 1
```

Or a record(Obs: the new record building syntax is so cool) 
Ex:

```
〉{x: 3, y: 100} | to toml
x = 3
y = 100
```

And finally a string that can be parsed as a toml(I thought this was a little bit weird given that we have `from toml` but i decided to keep it to keep the compatiblity)

```
〉'foo = 1' | to toml
foo = 1
```
It will output an error if a string that cant be parsed is given.

```
〉'hello world' | to toml
Error: nu::shell::unsupported_input (https://docs.rs/nu-protocol/0.1.0/nu-protocol/enum.ShellError.html#variant.UnsupportedInput)

  × Unsupported input
   ╭─[entry #7:1:1]
 1 │ 'hello world' | to toml
   · ──────┬──────
   ·       ╰── "hello world" unable to de-serialize string to TOML
   ╰────
```


jntrnr:
Looks good

PR #402, opened 2021-12-02
feat(random): add random-decimal

jaeheonji:
Porting decimal sub-command in random

Perhaps, there might be a better way? Please let me know if you have any!


jntrnr:
this looks good!

PR #404, opened 2021-12-02
Port str upcase

onthebridgetonowhere:


crates/nu-command/src/strings/str_/upcase.rs
```diff
@@ -0,0 +1,109 @@
...
+            }
+        },
+        engine_state.ctrlc.clone(),
+    )
+}
+
+fn action(input: &Value, head: Span) -> Value {
+    match input {
+        Value::String { val: s, .. } => Value::String {
+            val: s.to_ascii_uppercase(),
```
jntrnr, line 77: I wonder... should this be `to_uppercase` instead of `to_ascii_uppercase`? Nushell can work with utf-8 strings, and that'd let it also uppercase more non-English words if I understand the docs right

onthebridgetonowhere: good point. I think you're right, we probably should use `to_uppercase`.

jntrnr:
looks good

PR #419, opened 2021-12-03
port over the reject command from nushell

stormasm:
completed porting reject...


jntrnr:
Nice. Can you also add some tests either as examples or to src/tests.rs?

src/tests.rs
```diff
@@ -1187,3 +1187,24 @@ fn comment_skipping_2() -> TestResult {
         "40",
     )
 }
+
+#[test]
+fn command_filter_reject_1() -> TestResult {
+    run_test("[[lang, gems]; [nu, 100]] | reject gems", "{lang: nu}")
```
jntrnr, line 1193: Hmm, this feels funny to me. `reject` would remove the "gems" column, but here it's also changing the type from table to record.

jntrnr: It's possible I'm misreading the test though

stormasm:
```rust
#[test]
fn command_filter_reject_4() -> TestResult {
    run_test(
        "[[lang, gems, grade]; [nu, 100, a]] | reject gems | to json",
        "[{lang:nu,grade:a}]",
    )
}
```
@jntrnr I tried this one as well inside the tests.rs to no avail...

PR #440, opened 2021-12-05
feat(random): add random-integer and random-uuid

jaeheonji:
Porting integer sub-command in random

Finally! This is the last random command on the porting list.

It is not currently on the porting list, but there is a [uuid](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/commands/random/uuid.rs) in the sub-command of random.
Would it be better to port this too? If it does, it would be fine to do it with this PR because the amount of code is very small.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1496,6 +1496,7 @@ dependencies = [
  "toml",
  "trash",
  "unicode-segmentation",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


fdncred:
@jaeheonji yes, we would like 100% feature parity with nushell. So, yes, if `random uuid` is missing we'd like to port it.

jaeheonji:
Hi @fdncred ! Thanks for answering my question.
Since the implementation of uuid is simple, I pushed a commit to this PR. :)

jntrnr:
Looks good!

PR #442, opened 2021-12-05
Batch of dataframe commands

elferherrera:


jntrnr:
more dataframe commands! we're getting close 😄

PR #443, opened 2021-12-06
introducing `gstat`, a new command to get the git status

fdncred:
the purpose of this command is to get the git status of a local repo in order to parse that status and use in a prompt string. the `wt` in the data stands for `working tree`. the `idx` in the data stands for `index`. the rest should be self-explanatory. since we don't really want this to fail when run outside of a git tracked folder, we just return -1 on the numeric fields so these could be checked from a script.

from inside a git tracked folder
image

from outside a git tracked folder
image

pass folder in on the command line
image

pipe folder/string in on the command line
image

TODO:
- [ ] fix any bugs found
- [ ] i was unable to pipe in bare words, i'm not sure why - maybe it's supposed to be that way


elferherrera:
I would add a category to the signature

fdncred:
Good point. It should be required.

PR #447, opened 2021-12-06
Port uniq

onthebridgetonowhere:
I switched from the original nushell `IndexMap` to tuples. If we have any performance issues on large inputs, we should maybe rethink it, but for now it seems to work fine.


jntrnr:
Looks good!

PR #452, opened 2021-12-08
Port `network/url` command

jaeheonji:
Porting url commands!

Because url commands use generic `operator` function, I didn't create sub-commands with different PRs.


fdncred:
Thanks!

PR #457, opened 2021-12-09
Dataframe commands name

elferherrera:
Change names in dataframe commands to use the word `dataframe`


fdncred:
my good friend dataframe is back!

PR #462, opened 2021-12-10
Add zip-support to extra

luccasmmg:
As @fdncred pointed out in my last PR related to the port of `to-md` and `to-html`, the zip support that i added only worked if you did `cargo run --all-features` now it works also if you use `cargo run --features=extra`.


fdncred:
Thanks!

PR #464, opened 2021-12-10
Port `hash`, `hash md5` and `hash sha256` commands

CBenoit:
Related to https://github.com/nushell/engine-q/issues/242.

`hash` by itself is only printing a help message.

The other two are using the same generic implementation so I figured it might as well be in the same PR.

`hash base64` command will be ported in a separate PR to simplify reviewing.


fdncred:
looks good. thanks!

PR #467, opened 2021-12-11
Port version

onthebridgetonowhere:
Not sure the `installed_plugins` part works correctly. I tried to install the dummy example plugin, but it either does not get loaded or something else is wrong.


fdncred:
right now there only 3 plugins with 5 commands. you can see them by doing this `$scope.commands | where is_plugin == $true`. in order to see plugins at all you have to:
1. compile with `cargo build --all-features` i still think `cargo build --features=extra` is broken
2. after that compile you have to do `register target/debug/nu_plugin_gstat` to register your plugin and be able to run it.
3. do the same with `nu_plugin_example` and `nu_plugin_inc`. `example` has `test-1`, `test-2`, and `test-3` commands.

after that you should be able to see all the plugins.

onthebridgetonowhere:
Many thanks @fdncred - I now managed to run the `test-1` command from the example plugin and I can debug my `version` command port.

fdncred:
are you still working on this or is it ready to land?

onthebridgetonowhere:
I forgot to delete some code and still did not test if the reported features are correct. I'll revert to a draft. Sorry about that!

onthebridgetonowhere:
@fdncred now it's ready for review, thanks!

crates/nu-command/src/core_commands/version.rs
```diff
@@ -0,0 +1,369 @@
...
+    engine_state: &EngineState,
+    _stack: &mut Stack,
+    call: &Call,
+    _input: PipelineData,
+) -> Result<PipelineData, ShellError> {
+    let tag = call.head;
+
+    let mut indexmap = IndexMap::with_capacity(4);
+
+    indexmap.insert(
```
fdncred, line 55: I'm wondering if this use of `CARGO_PKG_VERSION` is right considering this https://github.com/baoyachi/shadow-rs#4-done-use-shadow. Let's just run with it and see what happens.

fdncred:
Thanks.

PR #479, opened 2021-12-12
Add in auto-cd if you pass just a directory

jntrnr:
Now:

```
> ..
```

Will automatically change directories. Auto-cd requires to start with `.`, `/`, or `\`.  The string thing makes this a little awkward for directories with spaces. Might need to think about that a bit (strings would just be an expression in e-q)

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -2,6 +2,7 @@ use std::borrow::Cow;
 use std::collections::HashMap;
 use std::env;
 use std::io::{BufRead, BufReader, Write};
+use std::path::Path;
 use std::process::{Command as CommandSys, Stdio};
 use std::sync::atomic::Ordering;
 use std::sync::mpsc;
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/system/run_external.rs
```diff
@@ -47,13 +48,31 @@ impl Command for External {
...
+        if (name.item.starts_with('.') || name.item.starts_with('/') || name.item.starts_with('\\'))
+            && path.is_dir()
+            && args.is_empty()
+        {
+            // We have an auto-cd
+            let _ = std::env::set_current_dir(&path);
+
+            //FIXME: this only changes the current scope, but instead this environment variable
+            //should probably be a block that loads the information from the state in the overlay
+            stack.add_env_var("PWD".into(), name.item.clone());
```
fdncred, line 72: there's `$nu.cwd` that gets updated as well now. i'm not sure if that needs to be updated here or if it's updated by default.

jntrnr: yeah, we may want to file an issue for cwd/pwd work. There are a couple rough edges and once we figure them out we should fix both `cd` and auto-cd

Issue #459, opened 2021-12-09
List of all configurable settings


kubouch:
Let's keep a list of what can be set up in engine-q until we document it properly.

## `$config`

```
let $config = {
  filesize_metric: $true
  table_mode: rounded # basic, compact, compact_double, light, thin, with_love, rounded, reinforced, heavy, none, other
  use_ls_colors: $true
  color_config: {
    separator: yd
    leading_trailing_space_bg: white
    header: cb
    date: pu
    filesize: ub
    row_index: yb

    bool: red
    int: green
    duration: red
    range: red
    float: red
    string: red
    nothing: red
    binary: red
    cellpath: red
  }
  use_grid_icons: $true
  footer_mode: always #always, never, number_of_rows, auto
  animate_prompt: $false
  float_precision: 2
  use_ansi_coloring: $true
  filesize_format: "b" # b, kb, kib, mb, mib, gb, gib, tb, tib, pb, pib, eb, eib, zb, zib, auto
}
```

## `$nu.env`

* `LS_COLORS`: Sets up colors per file type in `ls`
* `PROMPT_COMMAND`: Code to execute for setting up the prompt

jntrnr:
We also need a way to configure the prompt indicators, so that it's easier to use things like starship

fdncred:
@jntrnr I added the prompt indicators from reedline above, is this what you are talking about?

jntrnr:
@fdncred - ah okay, so those aren't yet supported. Cool.

Oh unrelated but while I'm here - I was wondering if we should pick nicer names for the syntax highlighting. flatshape_int feels kind of like you have to know how nushell works internally to know what that means :D Maybe we could use "nu_int" or "lang_int" or something?

fdncred:
I could go for `cli_int` since it's really coloring the cli syntax - or `syntax_int`.

fdncred:
The only items that are the same between flatshape and primitives are `int`, `float`, `range`, `string` - all there rest are different and not included in the primitive coloring. Also, the flatshapes have different coloring that primitives. so, I think we need to keep the flatshapes different from primitives as far as coloring. I was once thinking about combining these, for instance, only have one `int` coloring, but now I think that would be wrong.

jntrnr:
I added `log_level` which lets you turn on tracing if you want. It currently restarting the shell when you change it (but maybe that's fixable with a few tweaks)

PR #489, opened 2021-12-13
allow fg, bg, attributes to be set for all colors in color_config

fdncred:
this change allows all colors in the color_config section of the config.nu to have foreground, background, and color attributes. the foreground and background are written in hex colors like #rrggbb and the attributes are a concatenated string with the following values and meaning.
```rust
    for ch in color_attr.to_lowercase().chars() {
        match ch {
            'l' => blink = true,
            'b' => bold = true,
            'd' => dimmed = true,
            'h' => hidden = true,
            'i' => italic = true,
            'r' => reverse = true,
            's' => strikethrough = true,
            'u' => underline = true,
            'n' => (),
            _ => (),
        }
    }
```
the config.nu color_config section could look like this. comments are allowed, quotes around the hex colors and attributes are required.
```
  color_config: {
    separator: "#88b719" #yd
    leading_trailing_space_bg: white
    # header: cb
    header: {
      fg: "#B01455",
      bg: "#ffb900",
      attr: bli
    }
    date: "#75507B" #pu
    filesize: "#729fcf" #ub
    # row_index: yb
    row_index: {
      fg: "#e50914"
    }

```
so the `bli` in the header attr section means bold blink italic.
let the crazy colors begin. :)
crazy2

### DIFF
--- a/crates/nu-protocol/src/config.rs
+++ b/crates/nu-protocol/src/config.rs
@@ -75,7 +75,7 @@ impl Value {
                                 span: _,
                             } => {
                                 // make a string from our config.color_config section that
-                                // looks like this: { fg: "#rrggbb", bg: "#rrggbb", attr: "abc", }
+                                // looks like this: { fg: "#rrggbb" bg: "#rrggbb" attr: "abc", }
                                 // the real key here was to have quotes around the values but not
                                 // require them around the keys.
 
@@ -87,7 +87,7 @@ impl Value {
                                     .zip(inner_vals)
                                     .map(|(x, y)| {
                                         let clony = y.clone();
-                                        format!("{}: \"{}\", ", x, clony.into_string(", ", &config))
+                                        format!("{}: \"{}\" ", x, clony.into_string(", ", &config))
                                     })
                                     .collect();
                                 // now insert the braces at the front and the back to fake the json string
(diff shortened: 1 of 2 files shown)


kubouch:
@fdncred Can you also update https://github.com/nushell/engine-q/issues/459 with the new config?

crates/nu-protocol/src/config.rs
```diff
@@ -65,10 +65,40 @@ impl Value {
...
+                                    .zip(inner_vals)
+                                    .map(|(x, y)| {
+                                        let clony = y.clone();
+                                        format!("{}: \"{}\", ", x, clony.into_string(", ", &config))
+                                    })
+                                    .collect();
+                                // now insert the braces at the front and the back to fake the json string
+                                val.insert(0, '{');
+                                val.push('}');
+                                hm.insert(key, val);
```
fdncred, line 96: is there a better way to do all this fake json manipulation with serde somehow? a more idiomatic way to deserialize this particular value to regular json?

fdncred:
> @fdncred Can you also update #459 with the new config?

ya, once I land this I will. I'd like to get some input on the section I talked about above.

crates/nu-protocol/src/config.rs
kubouch: Do you need to create a JSON string here? Could you create NuStyle directly from the Record?

fdncred: I think it needs to look like the json in the comments above this section in order to parse it later since this section is building a hashmap of string, value.

fdncred: i think we'll just run with this and see what happens

PR #493, opened 2021-12-14
Fix error propagation across `hash` commands

CBenoit:
Example:

```
ls | get name | hash sha256
```

### DIFF
--- a/crates/nu-command/src/hash/generic_digest.rs
+++ b/crates/nu-command/src/hash/generic_digest.rs
@@ -1,7 +1,7 @@
 use nu_engine::CallExt;
 use nu_protocol::ast::{Call, CellPath};
 use nu_protocol::engine::{Command, EngineState, Stack};
-use nu_protocol::{Example, PipelineData, ShellError, Signature, Span, SyntaxShape, Value};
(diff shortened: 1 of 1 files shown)


fdncred:
I'm confused. What does this do exactly? This already works for me on Windows and Linux.
image

CBenoit:
Oh. That's my bad, I originally tried the following and went wondering why it didn't work:

image

With this PR it's not getting any better as the output of cat is not getting piped as expected.

image

I guess all we need is to wait for `open` to be supported instead.

Let me repurpose this PR. I was originally just fixing the span / error when receiving an error value as input.

Before:
image

After:
image

fdncred:
how about this `ls | where type == file | each {|r| (cat $r.name) | hash md5 }`

CBenoit:
> how about this `ls | where type == file | each {|r| (cat $r.name) | hash md5 }`

Oh! Indeed, it works perfectly :ok_hand: 

And I push-forced to keep only the error propagation fix.

fdncred:
oh, when the ci finishes we can land this.

jntrnr:
looks good

PR #494, opened 2021-12-14
Porting 'ansi' command from nushell to engine-q

matthewauld:
I made a couple of changes - it seems in nushell you could provide an escape, osc, and default value, but only the first one would be actually performed, and the rest ignored. I changes escape and osc to mutually exclusive switches, and changes the optional positional arg to required. I think this was how it was intended to work - but if I am wrong, happy to change it back, just let me know!

I don't have a ton of formal rust experience, so any feedback appreciated.

### DIFF
--- a/crates/nu-command/src/platform/ansi/command.rs
+++ b/crates/nu-command/src/platform/ansi/command.rs
@@ -124,9 +124,9 @@ Format: #
         if escape && osc {
             return Err(ShellError::IncompatibleParameters {
                 left_message: "escape".into(),
-                left_span: call.get_named_arg("escape")?.span,
+                left_span: call.get_named_arg("escape").expect("Unexpected missing argument").span,
                 right_message: "osc".into(),
-                right_span: call.get_named_arg("osc")?.span,
+                right_span: call.get_named_arg("osc").expect("Unexpected missing argument").span,
             });
         }
         if escape || osc {
@@ -134,7 +134,7 @@ Format: #
             if code_vec[0] == '\\' {
                 return Err(ShellError::UnsupportedInput(
                     String::from("no need for escape characters"),
-                    call.get_flag_expr("escape")?.span,
+                    call.get_flag_expr("escape").expect("Unexpected missing argument").span,
                 ));
             }
         }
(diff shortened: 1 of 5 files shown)


jntrnr:
@matthewauld - nice! thanks for the PR

A couple things CI is going to check that you might want to run locally: `cargo test --all --all-features` and `cargo clippy --all --all-features`.

If both of those come back green and without warnings you should be good. The checked the code and it seems good. Once CI passes you should be good to land.

matthewauld:
Thanks - I didn't run all the tests it seems - Ill fix and update. Sorry about that!

matthewauld:
Alright - that should do it!

jntrnr:
@matthewauld - ahh oops, there's one last command I should have mentioned above. You'll need to run `cargo fmt` before you submit. This will give the code a standard formatting.

matthewauld:
Not a problem!

matthewauld:
Ah, I see - thanks!

fdncred:
Hey @matthewauld, thanks for this. 

I have a request for a follow up pr to this ansi command. As always, feel free to say no, but I figured I'd ask. 

I'd like to land this to get this base functionality so people can play with it but I'd really like it to work like `char` in Nushell works, meaning, I'd like to have the ability to do `ansi --list` and see all the currently available ansi commands.

Is that functionality something you'd be interested in adding?

crates/nu-command/src/platform/ansi/command.rs
```diff
@@ -0,0 +1,354 @@
...
+                "escape", // \x1b[
+                "escape sequence without the escape character(s)",
+                Some('e'),
+            )
+            .switch(
+                "osc", // \x1b]
+                "operating system command (ocs) escape sequence without the escape character(s)",
+                Some('o'),
+            )
+    }
```
fdncred, line 33: I think we also need a category but not sure which category to use. Just pick one please and run with it, we can always change it later.

matthewauld:
@fdncred  - sure, happy to! Should be able to get it done tonight

fdncred:
> @fdncred - sure, happy to! Should be able to get it done tonight

great! thanks so much!!!

do you mind adding a category to this before we land it?

matthewauld:
Yep, sorry about that!

jntrnr:
Thanks!

PR #497, opened 2021-12-14
Treating environment variables as Values

kubouch:
## Environment Variables as Values

This PR changes environment variables to be any Value instead of just a String. For example, PATH can be a list:
path_list2

## Configurable Translations from/to a String

You can define custom translation from String (captured from host environment) to Value and from Value back to String (to be sent to external command). For example, translation config for PATH can look like this:
```
let config = {
    # ... other config ...
    env_conversions: {
            "PATH": {
                from_string: { |s| $s | split row ':' }
                to_string: { |v| $v | str collect ':' }
            }
    }
}
```

## No Removing with Nothing

Also, I removed the Nothing value removing the environment variable. If you set an environment variable to Nothing, it will hold this value. If you want to remove it, use `hide`.

## `env` Command 

For easier inspection and debugging, there is now an `env` command:
env_command

## TODO (can be left for another PR)
- [x] Make PROMPT_COMMAND a block instead of a string (enables syntax highlighting when writing it)
    - [x] Fix prompt not updating as well
- [x] Check spans of the env var values  (we could have a fake source file that holds the ones captured from `std::env`)
- [x] Change BlockID to Value::Block in EnvConversions to have the span as well (it's (BlockId, Span) tuple now...)
- [x] Check how error messages look like
- [x] `env` command for convenient env var inspection. Ideas:
    - [x] (added column) `env raw` would print the translated values sent to external programs)
    - [ ] `env path` would print the PATH variable name (PATH on Unix, Path on Windows)
    - [x] (added column) `env list` would just print the env vars same as calling `$nu.env` but also print their types
    - [ ] `env to-string` and `env from-string` could encode/decode the env var, but that is the same as calling `do $config.env_conversions.X.from_string` so maybe not necessary
- [x] Move the translation code somewhere instead of free-floating
- [ ] Add default config for PATH etc.
- [ ] Add docs page about this
- [x] Fix env vars passed to external tools:
    - [x] LS_COLORS 
    - [x] run_external 
- [x] Verify externals get all the env
- [ ] Cache the translated strings alongside Values to avoid repeated eval
- [ ] Investigate shorthand `FOO= $nu.env.FOO` being string and not nothing
- [ ] Add tests
- [ ] Better description and examples for env command
- [ ] Add `Value::into_env_string()` -- same as `into_string()` but more env-friendly (e.g., use the env separator by default and do not put `[...]` around lists.

### DIFF
--- a/crates/nu-cli/src/prompt.rs
+++ b/crates/nu-cli/src/prompt.rs
@@ -8,7 +8,6 @@ use {
 /// Nushell prompt definition
 #[derive(Clone)]
 pub struct NushellPrompt {
-    prompt_command: String,
     prompt_string: String,
     // These are part of the struct definition in case we want to allow
     // further customization to the shell status
@@ -27,7 +26,6 @@ impl Default for NushellPrompt {
 impl NushellPrompt {
     pub fn new() -> NushellPrompt {
         NushellPrompt {
-            prompt_command: "".to_string(),
             prompt_string: "".to_string(),
             default_prompt_indicator: "〉".to_string(),
             default_vi_insert_prompt_indicator: ": ".to_string(),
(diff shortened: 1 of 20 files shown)


fdncred:
this is cool!

jntrnr:
Wow this is cool

PR #499, opened 2021-12-15
feat(into): add into-bool command

jaeheonji:
This PR is related nushell/nushell#4302

**TODO**
- [x] convert `decimal` to boolean
- [x] convert `decimal string` to boolean
- [x] convert `string` to boolean
- [x] convert value to boolean in table

**Expected behavior**
* decimal `0` or `0.0` (float type) convert to `false`, other decimal is `true`
  * (Considerations) Should `true` only allow decimal `1`?
* `decimal string` is same `decimal`. just string type.
* `string` can accept a `'true'` or `'false'` string.


crates/nu-command/src/conversions/into/bool.rs
```diff
@@ -0,0 +1,183 @@
...
+    fn name(&self) -> &str {
+        "into bool"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("into bool")
+            .rest(
+                "rest",
+                SyntaxShape::CellPath,
+                "column paths to convert to binary (for table input)",
```
onthebridgetonowhere, line 21: I think the comment should say `to boolean` instead of `to binary`

jaeheonji: Oh! thanks for finding my mistake 😅

jntrnr:
Looks good!

PR #500, opened 2021-12-15
Porting  'char' command from nushell to engine-q

matthewauld:
Nothing too exciting here. Checked with clippy, ran fmt. Thanks for your patience! #242


fdncred:
sweet - looks like my two favorite commands may be done today. good work  @matthewauld !!!

matthewauld:
Sorry about that, unit tests were depending on a fix from my other commit. Should run now. Apologies

jntrnr:
thanks!

PR #504, opened 2021-12-16
added a 'list' option to the ansi command

matthewauld:
Had to change the name -> code from a match statement to a vector that generates a hashmap, so it can be listed. Since there are a lot of shortcodes, I figured just a hashmap literal would get a little too big. Let me know if you can think of a more elegant solution - I am sure there is one.

I also wasn't sure how I should actually display the code - I just removed the escape character, but could print it as text if you prefer. 

Also added a category to the char command - I missed it in my last commit, figured I would sneak it in.


fdncred:
wow! that was fast! I can't wait to try this out but I'll be a couple hours.  thanks @matthewauld!!!

fdncred:
looks good to me. thanks

PR #507, opened 2021-12-16
Make dialoguer completion abortable

### LINKED ISSUE #505: Aborting `dialoguer` completions with `Ctrl-C` panics
## Steps to reproduce

1. Start typing
2. Enter tab completion with `Tab`
3. Press `Ctrl-C` to abort completion

image

sholderbach:
Fixes #505

Q: Are there any other error cases for dialoguer we have to handle separately?
A: Dialoguer returns a `std::io::Error`, thus if we cannot handle stdout anymore the reedline part would fail loudly as well

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -64,7 +64,7 @@ impl CompletionActionHandler for FuzzyCompletion {
                 .default(0)
                 .items(&selections[..])
                 .interact_on_opt(&Term::stdout())
-                .expect("Fuzzy completion interact on operation");
+                .unwrap_or(None);
             let _ = crossterm::terminal::enable_raw_mode();
 
             if let Some(result) = result {


fdncred:
good question. I"m not sure. I know we've also talked about using the technology from `ls | grid -c` somehow either with `dialoguer` or in place of it.

sholderbach:
We might want to pull in some of the display technology into reedline, this is probably just a quick fix to make sure my habits don't crash it

fdncred:
thanks. we could pull in `nu-grid` - you can just build and run that crate - I think it has a main.rs

sholderbach:
nu-term-grid definitely looks more concise but as far as I can see doesn't provide the selection option of dialoguer. but could be a good starting point

fdncred:
Ya, right. It's not the same type of app as dialoguer. It just creates a grid.

PR #512, opened 2021-12-17
Port flatten

onthebridgetonowhere:
This took a while. I tried to preserve nushell's original logic - which is fairly convoluted in my view, but don't know how to improve it. Feedback is more than welcomed, as always!


jntrnr:
looks good!

PR #515, opened 2021-12-17
Migration of series commands

elferherrera:
Dataframe commands to work with series


fdncred:
`df` should at least be `dfr` as to not conflict, imo.

elferherrera:
yeah sure. I can live with that

PR #516, opened 2021-12-18
Calling plugin without shell

elferherrera:
Now plugins shell out to cmd or sh only if the file extension is .bat or .sh
An option to shell to python was added, this allows for plugins in python to be developed 

A python plugin example was added. To register the example just use
```
register crates\nu-plugin-python\plugin -e json
```


fdncred:
let's try it!

PR #517, opened 2021-12-18
Plugin option for shell

elferherrera:
option to register a shell by indicating what shell can be used to call the plugin

```
register crates\nu_plugin_python\plugin.py -e json -s c:\Users\One\AppData\Local\Programs\Python\Python38\python.exe
```


fdncred:
yay, we're green. let's try it.

PR #519, opened 2021-12-18
add in a new command called columns

stormasm:
This gives us the ability to now get just the columns from the table,
as we discussed yesterday on discord...


fdncred:
oooooh, fun!

stormasm:
@fdncred yes ! and the name makes sense too :)

stormasm:
@fdncred ok cool let me know if you have any issues with how it works......

PR #527, opened 2021-12-19
nothing variable

elferherrera:
Introduces the variable `$nothing` and comparisons with other values


crates/nu-cli/src/syntax_highlight.rs
```diff
@@ -44,6 +44,11 @@ impl Highlighter for NuHighlighter {
                     get_shape_color(shape.1.to_string(), &self.config),
                     next_token,
                 )),
+                FlatShape::Nothing => output.push((
+                    // nushell Garbage
```
fdncred, line 48: still says garbage

crates/nu-color-config/src/shape_color.rs
```diff
@@ -3,143 +3,28 @@ use nu_ansi_term::{Color, Style};
...
-                lookup_ansi_color_style(int_color.to_string())
-            } else {
-                Style::new().bold()
-            }
-        }
-        _ => Style::default(),
+    match conf.color_config.get(shape.as_str()) {
+        Some(int_color) => lookup_ansi_color_style(int_color.to_string()),
+        None => match shape.as_ref() {
+            "flatshape_garbage" => Style::new().fg(Color::White).on(Color::Red).bold(),
```
fdncred, line 9: this is all definitely a lot cleaner.

crates/nu-color-config/src/shape_color.rs
```diff
@@ -3,143 +3,28 @@ use nu_ansi_term::{Color, Style};
...
-        "flatshape_custom" => {
-            if conf.color_config.contains_key("flatshape_custom") {
-                let int_color = &conf.color_config["flatshape_custom"];
-                lookup_ansi_color_style(int_color.to_string())
-            } else {
-                Style::new().bold()
-            }
-        }
-        _ => Style::default(),
+    match conf.color_config.get(shape.as_str()) {
```
fdncred, line 6: the reason i added `conf.color_config.contains_key["blah"]` is because `conf.color_config["blah"]` would puke if the key didn't exist. i guess `conf.color_config.get()` doesn't behave that way?

crates/nu-color-config/src/shape_color.rs
```diff
@@ -3,143 +3,28 @@ use nu_ansi_term::{Color, Style};
...
+            "flatshape_literal" => Style::new().fg(Color::Blue),
+            "flatshape_operator" => Style::new().fg(Color::Yellow),
+            "flatshape_signature" => Style::new().fg(Color::Green).bold(),
+            "flatshape_string" => Style::new().fg(Color::Green),
+            "flatshape_filepath" => Style::new().fg(Color::Cyan),
+            "flatshape_globpattern" => Style::new().fg(Color::Cyan).bold(),
+            "flatshape_variable" => Style::new().fg(Color::Purple),
+            "flatshape_flag" => Style::new().fg(Color::Blue).bold(),
+            "flatshape_custom" => Style::new().bold(),
+            _ => Style::default(),
```
fdncred, line 27: should there maybe be `"flatshape_nothing" => Style::default()` just to be complete?

crates/nu-color-config/src/shape_color.rs
elferherrera: It returns and option instead of error. That allowed me to clean most of the code

crates/nu-cli/src/syntax_highlight.rs
elferherrera: Le me correct the comment

crates/nu-color-config/src/shape_color.rs
fdncred: ah, ok. I didn't realize it returned an option. nice.

crates/nu-color-config/src/shape_color.rs
elferherrera: Yeap. Missing that one as well

jntrnr:
@elferherrera - you might want to add some tests to ensure that it's working

elferherrera:
> @elferherrera - you might want to add some tests to ensure that it's working

Tests added for parsing and value

PR #528, opened 2021-12-19
port empty command

onthebridgetonowhere:
The original `empty?` command was somewhat more complex. I honestly don't know if this is the right port for this command. Tests pass, though I feel that I did some things wrong, particularly with the `block` part. 
Would love some feedback on this or feel free to update it as you see fit. 

Also, would a `empty?` on a `Value::Nothing` type return `true` or `false`? What's the semantic here?


jntrnr:
This looks close to me. Let's go ahead and land it and then we can continue improving it if we find things we want to improve later on.

PR #532, opened 2021-12-20
allow `str index-of` and `str substring` to work with CJK

fdncred:
This was implemented to fix a problem in Nushell. Issue https://github.com/nushell/nushell/issues/4215.

This should work now.
```rust
let s = '[Rust 程序设计语言](title-page.md)'
let start = ($s | str index-of '(')
let end = ($s | str index-of ')')
echo ($s | str substring $'($start + 1),($end)')
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1709,6 +1709,7 @@ dependencies = [
  "rand",
  "rayon",
  "regex",
+ "ropey",
  "roxmltree",
  "rust-embed",
  "serde",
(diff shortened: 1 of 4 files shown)


jntrnr:
Hmm, I think we should think about what indices mean with strings. We're currently following along with what Rust does, which is to often use the byte index. For example:

```
〉let s = '[Rust 程序设计语言](title-page.md)'

〉$s | str length
40
```

I wonder - can we just fix `index-of` and `substring` to match what Rust does? I lean to yes, but if we'd rather go a different direction we do a bit of design first to understand what our basis will be for working with strings.

hustcer:
@jntrnr 
> We're currently following along with what Rust does, which is to often use the byte index.

IMO, When writing nushell it's just nushell not rust, the length should match the result of the colomn width displayed when we paste `[Rust 程序设计语言](title-page.md)` into VScode，as that's more intuitive.

fdncred:
In engine-q and Nushell this works and gives the right results.

```
> let s = "[Rust 程序设计语言](title-page.md)"
> $s | str length
40
```
40 bytes is right I think.
```
> $s | split chars | length
28
```
with this fix, the result of the lines above returns `title-page.md` without this fix, the result is `d)`

jntrnr:
@hustcer - unfortunately, the column width isn't reliable for strings in the terminal. One example would be something like:

🙅‍♂️

Which in some terminals will be one character and in other terminals may be multiple characters.

I agree that bytes feels a bit complicated, as it's not the first thing you'll think about when you're counting how many characters over to work with. I'm just worried that with some terminals counting characters also will give you a wrong answer.

fdncred:
closing in favor of standardizing on byte offsets as per #538

PR #534, opened 2021-12-20
Flatten should flatten embedded table

### LINKED ISSUE #533: to auto-flatten or not to auto-flatten
Nushell auto-flattens so you can do this command and get a list of every parameter.

```
> $nu.lang | get signature
```

With engine-q I have not figured out how to do this simple task. Here is a list of the ones I've tried.

`$scope.commands | get 0 | get signature` works for 1 command
`$scope.commands | get signature | flatten` doesn't work
`$scope.commands | each {|r| $r | get $r.signature}` doesn't work
`$scope.commands | each {|r| $r.signature}` doesn't work
`$scope.commands | select signature` doesn't work
`$scope.commands.signature | select command parameter_name parameter_type syntax_shape is_optional short_flag description` doesn't work
`$scope.commands | flatten signature` doesn't work
`scope.commands | each {|r| ($r | get signature).0 }` doesn't list all the params
`$scope.commands | each {|r| ($r | get signature) } | get 220` doesn't get all the commands, of course
`$scope.commands | each {|r| ($r | get signature) }` gets all of them but they're in tables and not flattened
`$scope.commands | each {|r| ($r | get signature) } | each {|s| $s | get parameter_name}` gets the parameter name but they're in a list and not flattened

imo, this needs to be fixed. it shouldn't be so difficult.

jntrnr:
Fixes #533 by opening `flatten` up to treat all lists the same. `flatten` now treats an inner table as a list of rows so that those rows can be flattened into the outer table/list.

### DIFF
--- a/crates/nu-command/src/filters/flatten.rs
+++ b/crates/nu-command/src/filters/flatten.rs
@@ -84,12 +84,12 @@ enum TableInside<'a> {
     Entries(&'a str, &'a Span, Vec<&'a Value>),
 }
 
-fn is_table(value: &Value) -> bool {
-    match value {
-        Value::List { vals, span: _ } => vals.iter().all(|f| f.as_record().is_ok()),
-        _ => false,
-    }
-}
+// fn is_table(value: &Value) -> bool {
+//     match value {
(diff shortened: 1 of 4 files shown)


fdncred:
much better! thanks!

PR #541, opened 2021-12-21
add `unknown` types to the `Type` enum

fdncred:
one thing this allows is the ability to see more type a parameters is in the `$scope.commands.signature | flatten`.


fdncred:
closing in favor of #545

PR #550, opened 2021-12-21
Interpret lists as series of args for externals

kubouch:
When you pass a list to an external, it will be treated as a series of arguments for the external:
```
> let l = [ a b c ]

> ^echo "foo" $l
foo a b c
```

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -50,7 +50,7 @@ impl Command for External {
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let mut name: Spanned<String> = call.req(engine_state, stack, 0)?;
-        let args: Vec<String> = call.rest(engine_state, stack, 1)?;
+        let args: Vec<Value> = call.rest(engine_state, stack, 1)?;
         let last_expression = call.has_flag("last_expression");
 
         // Translate environment variables from Values to Strings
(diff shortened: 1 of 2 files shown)


crates/nu-engine/src/eval.rs
```diff
@@ -470,12 +470,12 @@ pub fn eval_variable(
...
-        let env_columns: Vec<String> = env_vars.keys().map(|x| x.to_string()).collect();
-        let env_values: Vec<Value> = env_vars.values().cloned().collect();
+        let env_columns = env_vars.keys();
+        let env_values = env_vars.values();
 
         let mut pairs = env_columns
-            .into_iter()
-            .zip(env_values.into_iter())
+            .map(|x| x.to_string())
+            .zip(env_values.cloned())
```
elferherrera, line 478: Does it need clone?

crates/nu-command/src/system/run_external.rs
```diff
@@ -85,9 +85,34 @@ impl Command for External {
             return Ok(PipelineData::new(call.head));
         }
 
+        let mut args_strs = vec![];
```
elferherrera, line 88: I would suggest to use Vec::with_capacity

crates/nu-engine/src/eval.rs
kubouch: Yeah, I think you get iterator over `&Value` at first so you need a new copy.

crates/nu-command/src/system/run_external.rs
```diff
@@ -85,9 +85,34 @@ impl Command for External {
...
+
+        for arg in args {
+            if let Ok(s) = arg.as_string() {
+                args_strs.push(s);
+            } else if let Value::List { vals, .. } = arg {
+                // Interpret a list as a series of arguments
+                for val in vals {
+                    if let Ok(s) = val.as_string() {
+                        args_strs.push(s);
+                    } else {
```
elferherrera, line 98: Im wondering if this should be a recursive function

crates/nu-command/src/system/run_external.rs
kubouch: It might be hard to figure out the correct capacity beforehand. You'd need to first unwrap all the Values and at that point you can just collect them already.

crates/nu-command/src/system/run_external.rs
kubouch: Yeah, could be. It would have a side effect though that it would work for nested lists as well. It might be more desirable to support only flat lists? Since in case of a nested list the chance is you have an error in your arguments.

PR #554, opened 2021-12-22
Add environment variables doc page

kubouch:
Adds a page explaining environment variables as values, their config, etc.

Also removes leftover page in favor of nushell/nushell#4305


docs/Environment_Variables.md
```diff
@@ -0,0 +1,103 @@
...
+However, you can configure selected any environment variable to be converted to/from any value:
+
+```
+# config.nu
+
+let config = {
+    ... other config ...
+    env_conversions: {
+        FOO: {
+            from_string: {|s| $s | split row ':' }
```
fdncred, line 28: how about `(char seep)` instead of `:`

kubouch: I wanted to keep it simple, but maybe I should add a note there about these list-like env vars like PATH etc.

PR #555, opened 2021-12-22
add in a raw flag in the command to json

stormasm:
This gives us the ability to get back the json with out the table formatting
in the string...

```rust
[[a b]; [jim susie] [3 4]] | to json -r
```

will produce just the json unformatted...

```rust
[{"a":"jim","b":"susie"},{"a":3,"b":4}]
```

Prior to this addition @fdncred pointed out to me that this
was an alternative solution, which is kind of nice too, but just
a bit more verbose...

```rust
[[a b]; [jim susie] [3 4]] | to json | str find-replace '\n' '' -a | str trim -a
```

### DIFF
--- a/crates/nu-command/src/formats/to/json.rs
+++ b/crates/nu-command/src/formats/to/json.rs
@@ -13,7 +13,9 @@ impl Command for ToJson {
     }
 
     fn signature(&self) -> Signature {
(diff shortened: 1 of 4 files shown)


jntrnr:
looks good

PR #572, opened 2021-12-24
Wire hex viewing into a few more places

jntrnr:


fdncred:
ya, this is closer. at least I can do `cat target/debug/engine-q | first 100` and get a hex display. progress = good. just need to wire up the same with text files like `cat todo.md | into binary | first 100` and it should be nearly complete.

PR #587, opened 2021-12-26
Use only $nu.env.PWD for getting the current directory

kubouch:
The engine now uses **only** the PWD environment variable for determining the current directory. As a consequence, the PWD is now local for each scope instead of a global state (`std::env`) and thus is thread-safe. In other words, you can use `cd` in a `par-each` loop:
```
> do {
    let dirs = [
        test-pwd/d1
        test-pwd/d2
        test-pwd/d3
        test-pwd/d4
    ]

    $dirs | par-each { mkdir -s $it }
    $dirs | par-each { cd $it; touch foo.txt }

    ls test-pwd/*/*
}
# ... lists test-pwd/{d1,d2,d3,d4}/foo.txt
```

The `std::env` is updated every time we merge the delta into the engine state. This ensures parser and completions have up-to-date current directory. Writing to `std::env` during the delta merge should be safe since we don't do it in multiple threads.

## Checklist
* [ ] Fix plugins (gstat) -- need to pass them cwd
    * Partially circumvented by updating current dir in `std::env` each `merge_delta()`
    * Added a TODO note in the plugin code
* [x] Fix parser (cwd sync when merging delta)
* [x] Fix completions (cwd sync when merging delta)
* [x] Fix ls -- uses glob which uses std::env
* [x] Ensure PWD is always absolute path
    * perhaps add error about it in nu_engine::env::current_dir()
* [x] All instances of glob::glob() get absolute path
* [x] Fix performance: `benchmark { 1..100000 | each { let x = 4 } }` went from 200-300 ms to ~1s. 
* [x] Fix panic on `"/home"<tab>`
* [ ] Reedline uses `std::env::current_dir()` for default prompt (I guess that's OK)
* [x] _Seems_ like (emphasis on _seems_) we're not passing paths to any other external crates which could silently use `std::env::current_dir()`
    * I checked what commands we have and not many of them interact with paths in obvious way and those that do were fixed in this PR.\
* [x] `for` loop as well
* [x] check other looping commands for similar pattern
* [x] `im::HashMap` in engine state vs `HashMap` in the stack -- probably high cost of collecting env vars
    * Using `im::HashMap` seems to make it slightly slower.

### DIFF
--- a/crates/nu-command/src/example_test.rs
+++ b/crates/nu-command/src/example_test.rs
@@ -47,7 +47,8 @@ pub fn test_examples(cmd: impl Command + 'static) {
         working_set.render()
     };
 
-    let _ = engine_state.merge_delta(delta);
+    let cwd = std::env::current_dir().expect("Could not get current working directory.");
+    let _ = engine_state.merge_delta(delta, &cwd);
 
     for example in examples {
         // Skip tests that don't have results to compare to
@@ -67,7 +68,7 @@ pub fn test_examples(cmd: impl Command + 'static) {
             (output, working_set.render())
         };
 
-        let _ = engine_state.merge_delta(delta);
+        let _ = engine_state.merge_delta(delta, &cwd);
 
         let mut stack = Stack::new();
 
--- a/crates/nu-command/src/example_test.rs
+++ b/crates/nu-command/src/example_test.rs
@@ -7,6 +7,8 @@ use nu_protocol::{
     engine::{Command, EngineState, Stack, StateWorkingSet},
     PipelineData, Span, Value, CONFIG_VARIABLE_ID,
 };
+#[cfg(test)]
+use std::collections::HashMap;
 
 #[cfg(test)]
 use crate::To;
@@ -48,7 +50,7 @@ pub fn test_examples(cmd: impl Command + 'static) {
     };
 
     let cwd = std::env::current_dir().expect("Could not get current working directory.");
-    let _ = engine_state.merge_delta(delta, &cwd);
+    let _ = engine_state.merge_delta(delta, &cwd, HashMap::new());
 
     for example in examples {
         // Skip tests that don't have results to compare to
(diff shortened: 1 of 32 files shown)


crates/nu-command/src/example_test.rs
```diff
@@ -67,10 +68,27 @@ pub fn test_examples(cmd: impl Command + 'static) {
...
         };
 
-        let _ = engine_state.merge_delta(delta);
+        let _ = engine_state.merge_delta(delta, None, &cwd);
 
         let mut stack = Stack::new();
 
+        // Set up PWD
+        match std::env::current_dir() {
+            Ok(cwd) => stack.add_env_var(
```
jntrnr, line 77: can we reuse the `cwd` from above?

crates/nu-command/src/filesystem/mv.rs
```diff
@@ -50,7 +50,7 @@ impl Command for Mv {
         let interactive = call.has_flag("interactive");
         let force = call.has_flag("force");
 
-        let path: PathBuf = current_dir()?;
+        let path = PathBuf::from(current_dir(engine_state, stack)?);
```
jntrnr, line 53: one small change we might want to do is a helper method on `stack` that gets the current_dir as a PathBuf for you

jntrnr:
This looks good, just found a couple small changes that we could do in this PR or do in a later PR.

crates/nu-command/src/filesystem/mv.rs
kubouch: @jntrnr It would be nice, however, the `current_dir()` has a secret engine dependency: In case PWD env var is not a string, we need `eval_block()` to convert it to string and I didn't want to introduce this dependency to the protocol. We could, of course, require that PWD must always be a string. Then, we could move it to the stack. (I'd prefer having `current_dir()` return a PathBuf and `current_dir_str()`return a String since both are used frequently.)

jntrnr:
testing it now

jntrnr:
A couple things:

* doing auto-cd with things like `..` don't seem to work
* completions are still based off the old directory, and don't seem to use PWD

PR #595, opened 2021-12-27
add ability to specify an ansi style

fdncred:
This adds the ability to specify an ansi style on the command line that will change the foreground, background, and attributes all at the same time.

Added this example to demonstrate.
```rust
Example {
    description: "Use ansi to color text with a style",
    example: r#"$"(ansi -e { fg: '#0000ff' bg: '#ff0000' attr: b })Hello Nu World(ansi reset)""#,
    result: Some(Value::test_string(
        "\u{1b}[1;48;2;255;0;0;38;2;0;0;255mHello Nu World\u{1b}[0m",
    )),
},
```
Screen Shot 2021-12-26 at 9 27 21 PM

### DIFF
--- a/crates/nu-color-config/src/nu_style.rs
+++ b/crates/nu-color-config/src/nu_style.rs
@@ -2,13 +2,13 @@ use nu_ansi_term::{Color, Style};
 use serde::Deserialize;
 
 #[derive(Deserialize, PartialEq, Debug)]
-struct NuStyle {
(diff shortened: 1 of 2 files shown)


fdncred:
restarted several times to get ci to finish, twice it finished without ubuntu finishing, this time it finished without windows finishing. so I'm counting that as good.

Issue #559, opened 2021-12-23
`to json -r` serializes datetime without spaces


fdncred:
`ls | to json -r`
```json
{"name":".git","type":"dir","size":480,"modified":"2021-12-2302:09:48.927406137+00:00"}
```
versus
`ls | to json`
```json
  {
    "name": ".git",
    "type": "dir",
    "size": 480,
    "modified": "2021-12-23 02:09:48.927406137 +00:00"
  },
```

note that there are no spaced between date and time and between time and timezone.

it looks like this is squeezing out all the whitespaces

https://github.com/nushell/engine-q/blob/ef59b4aa5124974ec1b0ec9d48e3a81d1fd22271/crates/nu-json/src/ser.rs#L1036

I think the values should be unchanged, right?

/cc @stormasm

stormasm:
@fdncred yes you are right this is not what we want --- let me figure out the next step on how to fix this --- basically we want to retain all of the white space inside the strings of the keys and values ---- but get rid of the white space everywhere else...  Good catch, I was so focused on my table syntax example that I obviously completely missed this ---- let me do some work ---- and then run it by you...

stormasm:
@fdncred so my PR #596 I think fixed this issue...
After you confirm this works the way you think it should we can close this issue (hopefully) or figure out the next step...  Thanks...

fdncred:
hey @stormasm, everything seems to work now. here's my test one-liner that I was breaking it with earlier. `ls | to json -r | from json | update size {|s| $s.size | into filesize} | update modified {|d| $d.modified | into datetime}`.

good work. closing.

stormasm:
@fdncred thank you for reviewing !  glad that it is working...

PR #596, opened 2021-12-27
fix issue #559: to json -r serializes datetime without spaces

stormasm:
Prior to this fix we were removing whitespace from inside the keys and values
as well as outside the keys and values....

Now we NO longer remove whitespace from inside the keys and values,
but instead are leaving them intact....


stormasm:
This should fix #559

jntrnr:
nice

PR #598, opened 2021-12-27
Handle external redirects better

### LINKED ISSUE #597: Strange behavior of external commands in `let` statements
When calling `let x = <external>`, it does not seem to capture its value. Furthermore, the string decoding is a bit unexpected.

```
> let p = char prompt

> $p
▶

> let x = pwd  # prints out pwd directly
/home/foo/bar 

> $x  # prints nothing


> $x | describe
binary 

> $x | decode utf-8  # prints nothing


> $x | decode utf-8 | describe
string

> let x = (pwd)  # does not print anything

> $x | describe  # how come this is a string and not a binary?
string
```

I'm not sure how much of this is an expected behavior. Maybe `let x = pwd` should capture the output and either automatically decode it or remove auto-decoding from `(pwd)`. `let x = pwd` and `let x = (pwd)` should probably behave the same.

jntrnr:
fixes #597

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -480,7 +480,7 @@ pub fn eval_subexpression(
 ) -> Result<PipelineData, ShellError> {
     for stmt in block.stmts.iter() {
         if let Statement::Pipeline(pipeline) = stmt {
-            for (i, elem) in pipeline.expressions.iter().enumerate() {
+            for elem in pipeline.expressions.iter() {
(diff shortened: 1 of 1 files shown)


fdncred:
I had to restart the stupid CI again. Not sure what's going on with CI lately.

fdncred:
ok, I'm not restarting anymore. Mac is just stalling out. I'm not sure why ci is borked today. since this passed ubuntu and windows we should probably just land it.

PR #605, opened 2021-12-27
create history file if it doesnt exit

elferherrera:
If there is no history file, e-q never creates one and the history is not stored


jntrnr:
yay!

PR #614, opened 2021-12-28
Don't read config in a tight loop

jntrnr:
When checking if we need to output to the screen, we had accidentally been checking the config at each loop step as it was the first step of `eval_block`. Instead, check the config closer to where the output will actually happen so we only pay for the lookup when we're already going to pay for an i/o.

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -384,7 +384,6 @@ pub fn eval_block(
     block: &Block,
     mut input: PipelineData,
 ) -> Result<PipelineData, ShellError> {
-    let config = stack.get_config().unwrap_or_default();
     let num_stmts = block.stmts.len();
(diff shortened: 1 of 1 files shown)


fdncred:
before this pr on my Mac
```
〉benchmark { 1..100000 | each { $it + 10 } | length }
5sec 331ms 467us 750ns
```
after
```
〉benchmark { 1..100000 | each { $it + 10 } | length }
164ms 332us 125ns
```

jntrnr:
@fdncred what does nushell do with `benchmark { echo 1..100000 | each { $it + 10 } | length }` on your mac?

fdncred:
@jntrnr 
```
benchmark { echo 1..100000 | each { $it + 10 } | length }
╭───┬──────────────────╮
│ # │ real time        │
├───┼──────────────────┤
│ 0 │ 72ms 565us 250ns │
╰───┴──────────────────╯
```

PR #617, opened 2021-12-29
port over from nushell the column flag for the length command

stormasm:
enable getting the number of columns from the length command
similar to nushell...

```rust
[[name,age,grade]; [bill,20,a] [a b c]] | length -c
```

prior to this we only had the length command for rows


crates/nu-command/src/filters/length.rs
```diff
@@ -15,27 +18,95 @@ impl Command for Length {
...
+        .into_pipeline_data()),
+        _ => Ok(Value::Int {
+            val: input.into_iter().count() as i64,
+            span: call.head,
+        }
+        .into_pipeline_data()),
+    }
+}
+
+fn getcol(
```
jntrnr, line 70: You might want to look into `get_columns` which is currently in the table viewer. The reason I mention it is that the way we draw tables now, it's possible that a column missing in the first row might get added by a later row.

We could probably move `get_columns` to somewhere a bit more central so multiple commands could take advantage of it, too.

stormasm: yes, agreed, let me review the best way to move forward.  thanks for your feedback, I was not happy with the current implementation...

stormasm:
@jntrnr ok thanks for your feedback...  Now that  PR #628 to move get_columns to a more central location has landed.   I went back in and re-factored this PR to use the new centrally located get_columns method...

jntrnr:
looks good

PR #621, opened 2021-12-29
Ported `ignore` command to engine-q

hustcer:
Any feedback on this is welcomed or feel free to update it as you see fit.


jntrnr:
Looks good!

PR #626, opened 2021-12-30
Port the `every` command

Pantonshire:
I wasn't completely sure how to handle negative strides; I'm silently casting them to `usize` because this is what the `nth` command does, but perhaps returning an error would be more appropriate?


jntrnr:
It's probably good for now - we'll probably circle back later and keep improving the errors as we go.

@Pantonshire - if you want, you could file an issue so we remember to go back and improve the error later

PR #628, opened 2021-12-31
move get_columns from the table_viewer to a central location

stormasm:
This is prep work for PR #617 

In this PR both the **table_viewer** and the **columns command** is using
the now centrally located **get_columns** method...

Once this PR lands we will then use this as well for the
column flag in the length command which is currently PR #617

### DIFF
--- a/crates/nu-command/src/filters/columns.rs
+++ b/crates/nu-command/src/filters/columns.rs
@@ -66,7 +66,7 @@ fn getcol(
             },
             ..,
         ) => {
-            let input_cols = get_input_cols(input_vals);
+            let input_cols = get_columns(&input_vals);
             Ok(input_cols
                 .into_iter()
                 .map(move |x| Value::String { val: x, span })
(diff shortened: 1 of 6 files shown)


stormasm:
@jntrnr I wasn't sure where to put the get_columns method.  I think it should go into nu-protocol but wasn't sure where... I guess we could create a util.rs file instead of column.rs but I thought maybe in the future other column type methods could go inside column.rs as well.  Open for suggestions...

jntrnr:
@stormasm - I was thinking maybe `nu-engine` might be a decent place for it. What do you think?

stormasm:
@jntrnr nu-engine sounds like a good spot to me.  now only question is where to place the method ---- seemed to me like column.rs works cause it tells the developer about what is in the file.  Otherwise util.rs turns into a dumping ground for random stuff; which was another option.  Again, open to suggestions...

jntrnr:
Looks fine to me

PR #637, opened 2022-01-01
Cleanup binary stream print a little

jntrnr:


fdncred:
isn't `skip` and `address_offset` the same thing? i.e. you want to skip over the first 100 bytes = I want to start printing at offset 100.

jntrnr:
@fdncred - oh, I thought that skip meant that you wanted to skip the first x bytes

PR #639, opened 2022-01-01
make prompt indicators configurable

fdncred:
This is supposed to allow prompt indicators to be configurable.

You should be able to set these env vars in your config.nu and e-q applies them.
```
let-env PROMPT_INDICATOR = "A "
let-env PROMPT_INDICATOR_VI_INSERT = "I "
let-env PROMPT_INDICATOR_VI_VISUAL = "V "
let-env PROMPT_MULTILINE_INDICATOR = "- "
```


jntrnr:
nice!

PR #640, opened 2022-01-01
Port fetch

onthebridgetonowhere:
There has been some ongoing discussion in discord about `fetch`. I think it's best to have some other pairs of eyes to review the code and get some feedback before it's ready for merging. 

We're not returning a stream (from what I understand), but instead we're returning values directly. We can change this as we need. 

Todo:

- [x] Timeout flag
- [x] ByteStream / Streams for ctrl+c


crates/nu-command/src/network/fetch.rs
```diff
@@ -0,0 +1,352 @@
...
+) -> Result<nu_protocol::PipelineData, nu_protocol::ShellError> {
+    let mut fetch_helper = Fetch::new();
+
+    fetch_helper.setup(engine_state, call, stack)?;
+
+    let runtime = tokio::runtime::Runtime::new()?;
+
+    let path = match fetch_helper.path {
+        Some(p) => p,
+        None => return Err(ShellError::UnsupportedInput("test".to_string(), call.head)),
```
onthebridgetonowhere, line 90: Must fix this before PR is merged.

jntrnr: I'll switch to draft so we don't do it accidentally

crates/nu-command/src/network/fetch.rs
```diff
@@ -0,0 +1,352 @@
...
+    };
+
+    runtime.block_on(fetch(
```
jntrnr, line 93: I was poking around the other day and saw this: https://github.com/seanmonstar/reqwest#blocking-client

Haven't had a chance to try it, but that might be a way we can use reqwest without also pulling in tokio and async/await into e-q. There's also https://docs.rs/reqwest/latest/reqwest/blocking/struct.Response.html#method.copy_to which might let us do the streaming. I noticed that `.bytes()` seems to grab the whole response, which for large downloads might take a while before we can put anything into the pipeline.

onthebridgetonowhere: We could do that, for sure. If we're not set on using `reqwest`, we could try `ureq` crate instead. It's a simpler and a bit lighter library than `reqwest`, though it does not support async and thus it does not require tokio. 

You would suggest to use `copy_to` for anything that requires to get the bytes back? Would that mean that we'd not convert the body of the request into a `value`, but instead use a stream and get the bytes from the request?

onthebridgetonowhere:
I wonder if we should not allow for "exiting" the command before it's finished, in case it takes too long or other issues. At the moment, if we do `fetch "https://test.com"` it will sort of hang for a few seconds and ctrl+c does not seem to work. Should we support this? 

Another question is if we should have a default timeout. I found cases where even a file seems to exist on a server, somehow it cannot be served. A default timeout would be useful in my view; any opinions?

jntrnr:
The way we've done it in a few other places is that we have the ByteStream or StringStream, and the stream can be broken with ctrl-c. I think this is probably best in case the user finds themselves on a slow network and wants to break out and try a different host to connect to.

You could do `copy_to` into an iterator/reader type thing, and have that work like a ByteStream, which is just an iterator that outputs vec of bytes.

crates/nu-command/src/network/fetch.rs
```diff
@@ -0,0 +1,386 @@
...
+        fetch_helper.url,
+        fetch_helper.has_raw,
+        fetch_helper.user.clone(),
+        fetch_helper.password,
+        fetch_helper.timeout,
+    ))
+}
+
+#[derive(Default)]
+pub struct Fetch {
```
jaeheonji, line 114: Hi! @onthebridgetonowhere 
I'm interested in fetch and post porting, so actually I've seen the original nushell code.

Personally, I wonder if the `Fetch` struct is really necessary.
This simply takes arguments, but I think it's too much for the purpose.

So, How about an approach like this? (I wrote it rough)
```rust
use use base64::encode;
...

struct Args {
  // url, has_raw, user ..
}

...

fn run_fetch( ... ) {
  let args: Args = {
    url: Some(call.req(engine_state, stack, 0)?);
    ...
  }

  ...

  fetch(engine_state, call, stack, args}
}
```

onthebridgetonowhere: Thanks for the suggestions @jaeheonji. I guess we can do that; it's sometimes confusing how to handle things as we have several styles going on at the moment, so it would be great to unify them (e.g., keep only the struct Args for different arguments). I mostly followed what was done in the original nushell fetch command and focused on getting it to work, without refactoring or simplifying the code. I am still working on the byte stream part, but feel free to suggest changes :-)

onthebridgetonowhere: I've been cleaning this up. What do you think @jaeheonji ?

onthebridgetonowhere:
@jntrnr I've rewritten the command to use ByteStreams. I am not sure I've done the right thing though, maybe you can have a look at it when you have some time and suggest how to improve? Particularly, the buffering part is bad... I don't know how to make it work without writing/reading to/from a file :=) The `copy_to` takes as a variable of a type that implements `Write`, so... I guess we cold use `Vec<u8>`, but then how do I rewrite the `BufferedReader` part to support `Vec<u8>`?

jntrnr:
@onthebridgetonowhere - taking a look now. If I can figure it out, I'll push an update

jntrnr:
@onthebridgetonowhere - couldn't figure out how to push to this branch, but here's the fixes I did to use the buf reader directly: https://github.com/nushell/engine-q/pull/660

crates/nu-command/src/network/fetch.rs
jaeheonji: LGMT!

jntrnr:
@onthebridgetonowhere - I'm going to go ahead and land what we have. Feel free to improve it more if you like, but I think what we have is a good starting point.

jntrnr:
merged as part of #660

onthebridgetonowhere:
Thanks for the help! The buffer part was easy, don't know how I did not figure it out :D

PR #652, opened 2022-01-03
Fix list and table print

### LINKED ISSUE #647: colour of table is slightly wrong
image

jntrnr:
fixes #647


crates/nu-parser/src/flatten.rs
```diff
@@ -44,6 +46,8 @@ impl Display for FlatShape {
             FlatShape::Signature => write!(f, "flatshape_signature"),
             FlatShape::String => write!(f, "flatshape_string"),
             FlatShape::StringInterpolation => write!(f, "flatshape_string_interpolation"),
+            FlatShape::List => write!(f, "flatshape_string_interpolation"),
```
fdncred, line 49: shouldn't list be `flatshape_list`? :)

Issue #661, opened 2022-01-04
Make `fetch` part of "extra"


jntrnr:
I think right now you need to do `--all-features` to get everything including `fetch`. Should make `fetch` part of extra.

onthebridgetonowhere:
I'll take a look at this tonight and do some cleanup for fetch.

onthebridgetonowhere:
We decided to make `fetch` a default command envisioning future uses of `fetch` for auto downloading things that we might need. We can probably close this issue when #664 is merged.

PR #664, opened 2022-01-04
Move fetch to extra and clean up some code

onthebridgetonowhere:
Moves fetch to extra (see #661) and cleans up some code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,6 @@ stable = ["default"]
 extra = [
 	"default",
 	"dataframe",
-	"fetch-support",
 	"gstat",
 	"zip-support",
 ]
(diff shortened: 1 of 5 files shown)


fdncred:
i'm conflicted about `fetch` being in extra. here's why.

reasons for moving it to extra
* we may want it in extra there because i think ssl is required and some platforms may have a problem with that.


reasons for not moving it to extra
* if we get to the place we want to be, which is recognizing when plugins are needed and automagically downloading them, we'll need `fetch` as a regular internal command.

onthebridgetonowhere:
I definitely agree with the second part, when we need to download things automatically. So how about making it always available? Now we're not usyng `async` + `tokio` anymore, so the only crate needed is `reqwest`...

jntrnr:
@fdncred are you thinking of making `fetch` part of the default features?

jntrnr:
Ahh yeah, +1 on making it part of the default commands.

fdncred:
ya, i think `default` is where we want to be with `fetch` to allow us to auto-download things in the future.

onthebridgetonowhere:
I made it to be a default command :-)

jntrnr:
thanks!

PR #677, opened 2022-01-05
allow sharing violation error to not fail completely

fdncred:
It allows the `ls` to continue but not have much information for the files in error.
image

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -131,9 +131,13 @@ impl Command for Ls {
                         Err(e) => {
                             if e.kind() == ErrorKind::PermissionDenied
                                 || e.kind() == ErrorKind::Other
+                            // on windows this is "The process cannot access the file
+                            // because it is being used by another process. (os error 32)"
+                            || e.raw_os_error() == Some(32)
                             {
                                 None
                             } else {
+                                eprintln!("raw {:?}", e.raw_os_error());
                                 return Some(Value::Error {
                                     error: ShellError::IOError(format!("{}", e)),
                                 });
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/filesystem/ls.rs
```diff
@@ -131,9 +131,13 @@ impl Command for Ls {
                         Err(e) => {
                             if e.kind() == ErrorKind::PermissionDenied
                                 || e.kind() == ErrorKind::Other
+                            // on windows this is "The process cannot access the file
+                            // because it is being used by another process. (os error 32)"
+                            || e.raw_os_error() == Some(32)
                             {
                                 None
                             } else {
+                                eprintln!("raw {:?}", e.raw_os_error());
```
jntrnr, line 140: we could make this part of the error we show the user rather than a separate println

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -131,9 +131,13 @@ impl Command for Ls {
                         Err(e) => {
                             if e.kind() == ErrorKind::PermissionDenied
                                 || e.kind() == ErrorKind::Other
+                            // on windows this is "The process cannot access the file
+                            // because it is being used by another process. (os error 32)"
+                            || e.raw_os_error() == Some(32)
```
jntrnr, line 136: I was wondering if we should just be more resilient here and always list the filename, regardless of the ErrorKind.

crates/nu-command/src/filesystem/ls.rs
fdncred: oops - didn't mean to leave that in

crates/nu-command/src/filesystem/ls.rs
fdncred: well, this part wants to return `metadata`. i looked at how to create it like `fs::Metadata { blah }` but didn't see anything like it. i'm fine with closing this for a better alternative. i was just looking for something to not fail.

fdncred:
closing in favor of #681

PR #678, opened 2022-01-05
Fixes how environment is cloned inside tight loops

kubouch:
The environment is cloned only if it has changed on the stack. If it hasn't changed (i.e., the stack is empty), no cloning is necessary. This prevents unnecessary cloning of the empty data structures as long as you don't modify the env vars in a tight loop.

This loop works now:
```
ls | where type == dir | each { cd $it.name; ls } | flatten
```

Also fixes how canonicalize error is displayed and a potential std::env creep in the `cd` command.

### DIFF
--- a/crates/nu-command/src/filesystem/cd.rs
+++ b/crates/nu-command/src/filesystem/cd.rs
@@ -37,7 +37,10 @@ impl Command for Cd {
                 let path = match nu_path::canonicalize_with(path, &cwd) {
                     Ok(p) => p,
                     Err(e) => {
-                        return Err(ShellError::DirectoryNotFoundHelp(v.span()?, format!("IO Error: {:?}", e)))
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/filesystem/cd.rs
```diff
@@ -34,11 +34,13 @@ impl Command for Cd {
         let (path, span) = match path_val {
             Some(v) => {
                 let path = v.as_path()?;
-                if !path.exists() {
-                    return Err(ShellError::DirectoryNotFound(v.span()?));
-                }
-
-                let path = nu_path::canonicalize_with(path, &cwd)?;
+                println!("cd: {:?}, cwd: {:?}", path, cwd);
```
jntrnr, line 37: I think you forgot a debug print

PR #701, opened 2022-01-07
add more chars

fdncred:
added a few more chars to help with prompt construction
```rust
        "nf-branch" => '\u{e0a0}'.to_string(),                     // 
        "nf-segment" => '\u{e0b0}'.to_string(),                    // 
        "nf-left-segment" => '\u{e0b0}'.to_string(),               // 
        "nf-left-segment-thin" => '\u{e0b1}'.to_string(),          // 
        "nf-right-segment" => '\u{e0b2}'.to_string(),              // 
        "nf-right-segment-thin" => '\u{e0b3}'.to_string(),         // 
        "nf-git" => '\u{f1d3}'.to_string(),                        // 
        "nf-git-branch" => "\u{e709}\u{e0a0}".to_string(),         // 
        "nf-folder1" => '\u{f07c}'.to_string(),                    // 
        "nf-folder2" => '\u{f115}'.to_string(),                    // 
        "nf-house1" => '\u{f015}'.to_string(),                     // 
        "nf-house2" => '\u{f7db}'.to_string(),                     // 
```

### DIFF
--- a/crates/nu-command/src/strings/char_.rs
+++ b/crates/nu-command/src/strings/char_.rs
@@ -76,6 +76,16 @@ lazy_static! {
         // Unicode can't be mixed with Ansi or it will break width calculation
         "branch" => '\u{e0a0}'.to_string(),                        // 
         "segment" => '\u{e0b0}'.to_string(),                       // 
+        "left_segment" => '\u{e0b0}'.to_string(),                  // 
+        "left_segment_thin" => '\u{e0b1}'.to_string(),             // 
+        "right_segment" => '\u{e0b2}'.to_string(),                 // 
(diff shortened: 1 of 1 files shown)


jntrnr:
should we prepend these with nerdfont?

fdncred:
we can but if we do that there will be more than these probably. probably most of the icons in the char list are nerdfonts I think.

jntrnr:
the reason I mention is that I *think* these aren't part of unicode. we might want to group the standard ones separately from the "add-on" chars from special fonts. just a thought

fdncred:
I compared the unicode code-point list to our characters and renamed the ones I could find with a `nf-`. I couldn't figure out what the weather ones were so I have no clue if they're unicode or not.

PR #709, opened 2022-01-09
Fix not equal returning error when same things are compared in some cases

### LINKED ISSUE #699: compare bool for inequality
seems odd that we can't do this. `if $true != $true { $true }`
image

i tried to `change bool to bool` but it didn't work - haha!

i guess i could say `if $true == $false` but that's no fun.

onsah:
## What
This PR fixes #699

## How
The equality operators now return
```rust
(Type::Bool, None),
```
so everything is allowed to be compared now. Which also fixes boolean inequality.

## Questions
While looking for this fix I found out that lots of code in operator type check is just replicated. I don't have much context but it looks like we can reduce duplication and avoid problems like being able to compare bools using `==` but not `!=`.

Also is there any reason why `Operator::Equal` doesn't have `(Type::Int, Type::Int)` and `(Type::Float, Type::Float)` patterns but `Operator::NotEqual` has them?

### DIFF
--- a/crates/nu-parser/src/type_check.rs
+++ b/crates/nu-parser/src/type_check.rs
@@ -280,6 +280,7 @@ pub fn math_result_type(
                 (Type::Duration, Type::Duration) => (Type::Bool, None),
                 (Type::Filesize, Type::Filesize) => (Type::Bool, None),
 
+                (x, y) if x == y => (Type::Bool, None),
                 (Type::Nothing, _) => (Type::Bool, None),
                 (_, Type::Nothing) => (Type::Bool, None),
                 (Type::Unknown, _) => (Type::Bool, None),
--- a/crates/nu-parser/src/type_check.rs
+++ b/crates/nu-parser/src/type_check.rs
@@ -247,58 +247,8 @@ pub fn math_result_type(
                     )
                 }
             },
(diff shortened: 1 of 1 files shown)


fdncred:
Generally speaking, yes, less duplicate code is better as long as it's still readable and understandable. Sometimes we may even avoid some idiomatic ways to do some things in rust because they're less readable.

I also think it's reasonable to have non/equality comparison to/from all types, some of which you've listed.

jntrnr:
I think for equals and not-equals, we should just tell the typechecker that any two values can be equality compared. In effect, it'd immediately become `(bool, None)` if we see `==` or `!=`

fdncred:
are we still waiting for something here?

jntrnr:
@fdncred - yeah, the typecheck for both equality and non-equality should just return (bool, None) in each case. Each only needs that result and not any of the other cases it's doing now.

onsah:
@jntrnr you mean also when types are different than each other? For example, `42 == "hello"` should type check right?

jntrnr:
@onsah - you got it

jntrnr:
thanks!

onsah:
@jntrnr Just wondering, do we want to change the evaluation so we return `$false` for something like `42 == "foo"` or we want to leave as is? (type check passes, evaluation fails with type error)

PR #731, opened 2022-01-13
Add instructions to run engine-q

sholderbach:
Prompted by a question on discord we might want to add infos for those not fully versed in the Rust ecosystem


jntrnr:
thanks!

fdncred:
The only problem here is that if you just clone the repo and do `cargo run --features=extra` you won't actually compile the plugins. I've found that you have to do `cargo build --features=extra` to build the plugins, strangely enough.

PR #767, opened 2022-01-17
Fix unicode word wrapping with ansi-cut

sholderbach:
Ansi-cut expects ranges of character numbers (of the non-ansi control
sequence characters) instead of byte indices.
This fixes the panics when wrapping of non-ASCII lines (which exceed
the demanded number of characters as byte indices).
Also rectifies some wrong wrapping of unicode containing lines that
don't panic


fdncred:
I've been wondering what was causing those ansi-cut panics.

jntrnr:
Thanks!

zhiburt:
Hi there 

> I've been wondering what was causing those ansi-cut panics.

It's assert :smile:

fdncred:
haha! ya, I saw those. :)

PR #768, opened 2022-01-17
Keybindings config

elferherrera:
Adds the ability to parse simple keybindings into the emacs editor

To add an event you need to add this to your config file
```
 keybindings: [
        {
          name: completion
           modifier: CONTROL
           keycode: Char_c
           mode: emacs
           event: {
             type: single
             event: Complete
           }
         }
         {
          name: menu                                                                                                                                                                                  modifier: CONTROL
           keycode: down
           mode: emacs
           event: {
             type: single
             event: ContextMenu
           }
         }
       ]
```

Note: There is a refactoring of main function to locate utility functions


src/fuzzy_completion.rs
```diff
@@ -0,0 +1,57 @@
+use dialoguer::{
```
fdncred, line 1: #766 just removed all this stuff I think. is there a reason we need it?

src/reedline_config.rs
```diff
@@ -0,0 +1,104 @@
...
+            ReedlineEvent::Edit(vec![EditCommand::InsertChar('p')]),
+            ReedlineEvent::Enter,
+        ]),
+    );
+
+    for keybinding in parsed_keybindings {
+        let modifier = match keybinding.modifier.as_str() {
+            "CONTROL" => KeyModifiers::CONTROL,
+            "SHIFT" => KeyModifiers::CONTROL,
+            _ => unimplemented!(),
```
fdncred, line 79: why only control and shift and not option/alt?

src/reedline_config.rs
```diff
@@ -0,0 +1,104 @@
...
+        ReedlineEvent::Multiple(vec![
+            ReedlineEvent::Edit(vec![EditCommand::InsertChar('p')]),
+            ReedlineEvent::Enter,
+        ]),
+    );
+
+    for keybinding in parsed_keybindings {
+        let modifier = match keybinding.modifier.as_str() {
+            "CONTROL" => KeyModifiers::CONTROL,
+            "SHIFT" => KeyModifiers::CONTROL,
```
fdncred, line 78: shouldn't this be `KeyModifiers::SHIFT`?

jmoore34:
Looks amazing with the record syntax, 10/10

src/fuzzy_completion.rs
elferherrera: None. I just didn't see the merge. I'll delete it

src/reedline_config.rs
elferherrera: Lazynes. I want to merge this before continuing

src/reedline_config.rs
elferherrera: You are right. I'll continue working on this

PR #781, opened 2022-01-20
update all cargo crates to edition 2021

stormasm:
it looks like on my local machine that the 2021 edition for all cargo crates is working....


fdncred:
nice!

PR #784, opened 2022-01-20
`PATH` for completions for each os

fdncred:
we used to have to check `Path` on Windows and `PATH` on linux/mac. It appears that we now can just check `PATH`, yay!


jntrnr:
Nice

PR #814, opened 2022-01-22
port sort_by without float (yet)

stormasm:
floats require Ord which is not yet implemented...

once this lands I may need some help on Ord.


stormasm:
Starting to review this as a possible solution for Ord

https://github.com/notriddle/rust-float-ord

PR #822, opened 2022-01-22
draft: Bump `ansi-cut` version to 0.2.0

zhiburt:
Hi there,

After overseeing #767.
I noticed an issue in `ansi-cut`.

If color was used like this `"\u{1b}[41m\u{1b}[30msomething\u{1b}[39m \u{1b}[34m123123\u{1b}[39m\u{1b}[49m"`
(A color was closed not each time it is used) we did not recognize that, therefore the colors could be screwed.

I did a fix so it must be handled right now.

And I am here point out about the issue and to bring the fix.

But doing the bump.
I did not found any tests.
Do you have any?

As it's not quite safe to make just a bump because now `ansi-cut` `panic`s when you index not a UTF-8 code point boundary (it uses byte indexing now as well)

____

I did not inspect deliberately these lines, just have a little concern that `word.cut(start_index..end_index+width)` must be used here?

https://github.com/nushell/engine-q/blob/310ecb79b670f04aaf52b6366a6fce2c130a5c75/crates/nu-table/src/wrap.rs#L155-L160

Thank you

### DIFF
--- a/crates/nu-table/Cargo.toml
+++ b/crates/nu-table/Cargo.toml
@@ -18,5 +18,5 @@ nu-protocol = { path = "../nu-protocol"}
 regex = "1.4"
 unicode-width = "0.1.8"
 strip-ansi-escapes = "0.1.1"
-ansi-cut = "0.1.1"
+ansi-cut = "0.2.0"
 atty = "0.2.14"
--- a/crates/nu-table/src/wrap.rs
+++ b/crates/nu-table/src/wrap.rs
@@ -152,7 +152,7 @@ fn split_word(cell_width: usize, word: &str) -> Vec<Subline> {
     let mut end_index;
 
     let word_no_ansi = strip_ansi(word);
-    for c in word_no_ansi.chars().enumerate() {
+    for c in word_no_ansi.char_indices() {
         if let Some(width) = c.1.width() {
             end_index = c.0;
             if current_width + width > cell_width {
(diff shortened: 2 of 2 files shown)


jntrnr:
I think we've been trying to find a good repro for this one.

Thanks for the bump!

zhiburt:
Aaaaaaa it's probably was a mistake to merge it so fast 😅 

that's why I've put a draft note.

I do hesitate a bit because of there's no tests which could spotlight the issues I've described :)

Or is this OK?

fdncred:
I don't understand this exactly. It directly reverses @sholderbach's fix that eliminated the ansi-cut errors. Are we saying v2.0 fixes those problems so now we need to go back to bytes iteration versus characters iteration?

zhiburt:
> I don't understand this exactly. It directly reverses @sholderbach's fix that eliminated the ansi-cut errors. Are we saying v2.0 fixes those problems so now we need to go back to bytes iteration versus characters iteration?

Back in the day (5 day ago) in #767 @sholderbach was right about your indexing.

Now I am saying there was an inner issue in `ansi-cut` (I've described it above).
And I am just updating the version with the fix.
But beside the `fix` `0.2.0` changes interface, so inadvertently it does revert @sholderbach changes.

But if you take a glance at my comment I am pointing out 2 questions there beside that.

jntrnr:
@zhiburt - oops, sorry for merging so fast. We can revert the PR for now until you think it's ready. Or we can just test what's landed and see how it works in practice

fdncred:
I don't think we have test.
I agree with your question in wrap.rs. We should look into that.

Thanks for your help. I've proposed to the team that we keep this commit and see what happens.

PR #831, opened 2022-01-23
Fix cd-ing into a file

### LINKED ISSUE #830: I can cd to a file
With the engine-q in main, I cd to a file, and I can ls it , 
image

onthebridgetonowhere:
Closes #830


kubouch:
Thanks! Looks good!

PR #832, opened 2022-01-23
capture keyboard event

fdncred:
<kbd>option</kbd> + <kbd>delete</kbd>
Screen Shot 2022-01-23 at 7 31 47 AM
<kbd>control</kbd> + <kdb>d</kbd>
Screen Shot 2022-01-23 at 7 34 05 AM

I can only capture one key event. Do you think we can make this loop somehow?
/cc @jntrnr 

Update: couldn't get nushell tables to work but got a raw mode to work.
Screen Shot 2022-01-23 at 3 47 49 PM

### DIFF
--- a/crates/nu-command/src/platform/input_keys.rs
+++ b/crates/nu-command/src/platform/input_keys.rs
@@ -1,5 +1,5 @@
-use core::time::Duration;
-use crossterm::{event::poll, event::Event, event::KeyCode, event::KeyEvent, terminal};
+use crossterm::QueueableCommand;
+use crossterm::{event::Event, event::KeyCode, event::KeyEvent, terminal};
 use nu_protocol::ast::Call;
 use nu_protocol::engine::{Command, EngineState, Stack};
 use nu_protocol::{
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/platform/input_keys.rs
```diff
@@ -0,0 +1,136 @@
...
+            example: "input-keys",
+            result: None,
+        }]
+    }
+}
+
+pub fn print_events() -> Result<Value, ShellError> {
+    stdout().flush()?;
+    terminal::enable_raw_mode()?;
+    let result = print_events_helper();
```
jntrnr, line 50: What happens if you put the loop around this?

jntrnr: (line 50 specifically)

fdncred: I guess I'm not sure how to do that. I did this and it doesn't print anything when I hit a key combination.
```rust
pub fn print_events() -> Result<Value, ShellError> {
    stdout().flush()?;
    terminal::enable_raw_mode()?;
    loop {
        let event = crossterm::event::read()?;
        if event == Event::Key(KeyCode::Esc.into()) {
            break;
        }

        print_events_helper(event)?;
    }
    terminal::disable_raw_mode()?;

    Ok(Value::nothing(Span::test_data()))
}
```

fdncred: took a different route.

PR #837, opened 2022-01-24
Add var vals and alias expansions to scope var

sophiajt:
```
> $scope | get aliases
╭───┬───────┬───────────╮
│ # │ alias │ expansion │
├───┼───────┼───────────┤
│ 0 │ ll    │ ls -l     │
╰───┴───────┴───────────╯
```
```
> $scope | get vars
╭───┬──────┬──────┬───────╮
│ # │ name │ type │ value │
├───┼──────┼──────┼───────┤
│ 0 │ $x   │ int  │   100 │
╰───┴──────┴──────┴───────╯
```

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -620,18 +620,29 @@ pub fn eval_variable(
         let mut output_cols = vec![];
         let mut output_vals = vec![];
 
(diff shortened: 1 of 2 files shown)


fdncred:
thanks for doing this. I think it adds a good degree of polish!

PR #846, opened 2022-01-25
History menu

elferherrera:
Introduces a history menu for engine q
it can be triggered with Ctrl+a

The menu presents the history in pages that can be navigated with down or up arrows (tab and shift tab)
to see the next page use ctrl+a and ctr + shift + a for previous page
All of these events can be changed in config file

To configure the menu in the config use
```
 history_config: {
   page_size: 10
   selector: ":"                                                                                                                          
   text_style: dark_gray
    selected_text_style: green_reverse
  }
```
the selector is a character that when typed with a number after it will select that input. e.g. `:4` will select that row

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2863,7 +2863,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.2.0"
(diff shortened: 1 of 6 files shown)


jmoore34:
Does this override `MoveToLineStart`?

jntrnr:
@jmoore34 - currently it does, but we're trying to find a better keybinding for it

PR #851, opened 2022-01-26
Start support for commandline args to nu itself

jntrnr:
# Description

This adds support to be able to pass options to the engine itself (like the `--stdin` flag from nushell). These flags must come before the script file name. After the script file name come the flags to the script.

I've wired in the help and signature system to give a decent help message similar to what an internal call would give.

So far I've only added one flag, but in theory we should be able to add more once we decide on the set we want to support.

# Tests

### DIFF
--- a/crates/nu-command/src/filesystem/mod.rs
+++ b/crates/nu-command/src/filesystem/mod.rs
@@ -14,7 +14,7 @@ pub use cp::Cp;
 pub use ls::Ls;
 pub use mkdir::Mkdir;
 pub use mv::Mv;
-pub use open::Open;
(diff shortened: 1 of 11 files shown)


fdncred:
This looks cool. There's a couple of args that I'd like to add eventually.

1. `--login` seems like this may be needed for chsh? can't remember exactly but remember there being a pr or two about it in nushell.
2. `--perf` - there's a noticeable slowdown in windows on launch that I'd like to perf and see what's causing it.

jntrnr:
@fdncred yeah we should definitely add `-l`/`--login`. It's practically required for a shell to support it as a lot of tools assume it's there.

Issue #850, opened 2022-01-26
Seems like the nothing value messes something up.


chen244:
# commands get
If a list has none value ,it cannot be get
![image](https://user-images.githubusercontent.com/60290287/151162604-e95b4ae0-5c99-45d3-953b-866164a30bae.png)
and first that list , it can work
![image](https://user-images.githubusercontent.com/60290287/151162653-6aa141eb-d150-404c-beb0-745564d4db40.png)

fdncred:
it would be good to have that `Cargo.toml` file too in order to test with it, if possible.

chen244:
> 

Its your cargo.toml , just this project,engine-q

chen244:
> it would be good to have that `Cargo.toml` file too in order to test with it, if possible.

Its your cargo.toml , just this project,engine-q

fdncred:
ok, I see it now. thanks! `open Cargo.toml | get bin`. Thanks!

fdncred:
ya, it must be something with the `$nothing` entry, like someone else said in discord.
```
> open ./Cargo.toml | get bin | flatten
╭───┬───────────────────┬───────────────────────────────────────┬───────────────────╮
│ # │       name        │                 path                  │ required-features │
├───┼───────────────────┼───────────────────────────────────────┼───────────────────┤
│ 0 │ nu_plugin_inc     │ src/plugins/nu_plugin_core_inc.rs     │ inc               │
│ 1 │ nu_plugin_example │ src/plugins/nu_plugin_core_example.rs │ example           │
│ 2 │ nu_plugin_gstat   │ src/plugins/nu_plugin_extra_gstat.rs  │ gstat             │
│ 3 │ engine-q          │ src/main.rs                           │ ❎                │
├───┼───────────────────┼───────────────────────────────────────┼───────────────────┤
│ # │       name        │                 path                  │ required-features │
╰───┴───────────────────┴───────────────────────────────────────┴───────────────────╯
```

```
> open ./Cargo.toml | get bin | flatten | get required-features 
Error: nu::shell::name_not_found (link)

  × Name not found
   ╭─[entry #15:1:1]
 1 │ open ./Cargo.toml | get bin | flatten | get required-features
   ·                                             ────────┬────────
   ·                                                     ╰── did you mean 'name'?
   ╰────
```

chen244:
Seems it has been solved. Then I close this issue

PR #860, opened 2022-01-27
Maybe solve the none bug?

chen244:
# Description

Try to solve https://github.com/nushell/engine-q/issues/850
  
# Tests

- [ ] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [ ] `cargo clippy --all --all-features -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- [ ] `cargo build; cargo test --all --all-features` to check that all the tests pass

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -618,25 +618,28 @@ impl Value {
                     Value::List { vals, span } => {
                         let mut output = vec![];
                         let mut hasvalue = false;
+                        let mut temp: Result<Value, ShellError> =
+                            Err(ShellError::IOError("can't follow stream paths".into()));
                         for val in vals {
-                            let temp = val.clone().follow_cell_path(&[PathMember::String {
+                            temp = val.clone().follow_cell_path(&[PathMember::String {
                                 val: column_name.clone(),
                                 span: *origin_span,
                             }]);
-                            if let Ok(result) = temp {
+                            if let Ok(result) = temp.clone() {
                                 hasvalue = true;
                                 output.push(result);
-                            } else if hasvalue {
-                                output.push(Value::Nothing{span:*span});
                             } else {
-                                return temp;
+                                output.push(Value::Nothing { span: *span });
                             }
                         }
-
-                        current = Value::List {
-                            vals: output,
-                            span: *span,
-                        };
+                        if hasvalue {
+                            current = Value::List {
+                                vals: output,
+                                span: *span,
+                            };
+                        } else {
+                            return temp;
+                        }
                     }
                     Value::CustomValue { val, .. } => {
                         current = val.follow_path_string(column_name.clone(), *origin_span)?;
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -619,7 +619,7 @@ impl Value {
                         let mut output = vec![];
                         let mut hasvalue = false;
                         let mut temp: Result<Value, ShellError> =
-                            Err(ShellError::IOError("can't follow stream paths".into()));
+                            Err(ShellError::NotFound(span));
                         for val in vals {
                             temp = val.clone().follow_cell_path(&[PathMember::String {
                                 val: column_name.clone(),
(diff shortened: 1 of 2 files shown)


crates/nu-protocol/src/value/mod.rs
```diff
@@ -617,18 +617,23 @@ impl Value {
...
-                            //     }
-                            // }
+                            }]);
+                            if let Ok(result) = temp {
+                                hasvalue = true;
+                                output.push(result);
+                            } else if hasvalue {
+                                output.push(Value::String {
+                                    val: "none".to_string(),
+                                    span: Span { start: 0, end: 0 },
```
jntrnr, line 632: this isn't going to quite work as it will skip the rows at the beginning that are missing the column.

I think what we tried in Nushell was to grab all the values out, and then check to make sure they weren't all empty. You could use a similar logic here and always send a `Value::Nothing` (don't do `Value::String` here because "none" is valid text from the user).

Then, at the end of the loop, we can check to see if we ever saw a real value. If we didn't, we error.

chen244: Now I check it at last, can you review it ? thanks

kubouch: @chen244 I tried your PR and it works fine on simple cases I tried.

You have one failing test but that test seems to be testing the bug you fixed. So instead of expecting it to fail you can set it to expect the correct result (column of 20 and $nothing).

crates/nu-protocol/src/value/mod.rs
```diff
@@ -617,24 +617,29 @@ impl Value {
                     }
                     Value::List { vals, span } => {
                         let mut output = vec![];
+                        let mut hasvalue = false;
+                        let mut temp: Result<Value, ShellError> =
+                            Err(ShellError::IOError("can't follow stream paths".into()));
```
kubouch, line 622: I would put here an error that has a span. Otherwise, if you hit this error, it will just print the message but won't show where the error happened. So you can pick an error type that has a span and pass the `orig_span` to it.

crates/nu-protocol/src/value/mod.rs
chen244: > @chen244 I tried your PR and it works fine on simple cases I tried.
> 
> You have one failing test but that test seems to be testing the bug you fixed. So instead of expecting it to fail you can set it to expect the correct result (column of 20 and $nothing).

Hello, I have used ShellError::NotFound now. But I have a question. The test is used to test missing_column_error. But My fix does not think it has missed but is a nothing. In the test, it does a "size" command which is like get.Should the test be changed to another one?

chen244: > @chen244 I tried your PR and it works fine on simple cases I tried.
> 
> You have one failing test but that test seems to be testing the bug you fixed. So instead of expecting it to fail you can set it to expect the correct result (column of 20 and $nothing).

Now I get it an empty column length , but I do not know if it is right

chen244: > > @chen244 I tried your PR and it works fine on simple cases I tried.
> > You have one failing test but that test seems to be testing the bug you fixed. So instead of expecting it to fail you can set it to expect the correct result (column of 20 and $nothing).
> 
> Now I give it an empty column length , but I do not know if it is right

kubouch: @chen244 You can change the test completely, the current one is not very helpful. You can remove it and add a new test that verifies you can select a column with a missing value. For example, `[ { name: ABC, size: 20 }, { name: HIJ } ].size.1 == $nothing`, use `run_test` instead of `fail_test` and check the return value for `true`.

chen244: > @chen244 You can change the test completely, the current one is not very helpful. You can remove it and add a new test that verifies you can select a column with a missing value. For example, `[ { name: ABC, size: 20 }, { name: HIJ } ].size.1 == $nothing`, use `run_test` instead of `fail_test` and check the return value for `true`.

I update the test now

kubouch:
I just updated the test's name and deleted a comment that I think was outdated. When the checks pass, I'll merge it.

Thanks a lot for the effort!

PR #877, opened 2022-01-28
Port rename

onthebridgetonowhere:
# Description

`Rename` command now works. I also added an option to change a specific column's name, e.g. `[[ a b]; [1 2]] | rename -c [a new_name]`. The spans also work nicely, pointing to the specified column's area if the specified column does not exist. 
If more columns are given than exist in the record, it skips the extra columns. 

# Tests


jntrnr:
looks good

PR #880, opened 2022-01-29
Port rotate

onthebridgetonowhere:
# Description

Adds the rotate command, including rotate counter-clockwise. I slightly changed the logic so it would be great to have some feedback. I also considered having a flag for counter-clockwise, such that we do not have so much duplicated code. Any thoughts on this one?
  
# Tests


crates/nu-command/src/filters/rotate/counter_clockwise.rs
```diff
@@ -0,0 +1,213 @@
...
+    for (idx, val) in old_column_names.iter().enumerate().rev() {
+        let mut res = vec![Value::String {
+            val: val.to_string(),
+            span: call.head,
+        }];
+
+        let new_vals = {
+            // move through the array every 2 elements, starting from our old column's index
+            // so if initial data was like this [[a b]; [1 2] [3 4]] - we basically iterate on this [1 2 3 4] array, so we pick 2, then 4, and then when idx decreases (notice the .rev()), we pick 1 and 3
+            for i in (idx..new_values.len()).step_by(2) {
```
jntrnr, line 183: should this be based on the size of the table? If you had a 3x3 table, or say a 2x3 table, will this also work?

crates/nu-command/src/filters/rotate/command.rs
```diff
@@ -0,0 +1,213 @@
...
+    let mut final_values = vec![];
+
+    // the number of initial columns will be our number of rows, so we iterate through that to get the new number of rows that we need to make
+    // as we're rotating clockwise, we're iterating from left to right
+    for (idx, val) in old_column_names.iter().enumerate() {
+        let new_vals = {
+            let mut res = vec![];
+            // move through the array every 2 elements, starting from our old column's index
+            // so if initial data was like this [[a b]; [1 2] [3 4]] - we basically iterate on this [3 4 1 2] array, so we pick 3, then 1, and then when idx increases, we pick 4 and 2
+            for i in (idx..new_values.len()).step_by(2) {
```
jntrnr, line 180: Ditto (lol, oops I did these in reverse, see below)

crates/nu-command/src/filters/rotate/counter_clockwise.rs
onthebridgetonowhere: good catch, let me see.

onthebridgetonowhere: ok, the step has to be `new_values.len() / total_rows`. I fixed it and pushing it now.

onthebridgetonowhere:
@jntrnr any opinions on having one command with a flag, instead of two separate commands? Main advantage is that there will be less duplicated code. Disadvantage is that users would need to switch from `rotate counter-clockwise` to `rotate -c` or `rotate --counter-clockwise`.

fdncred:
i'm not sure what jt will say but i'd vote for having 1 command instead of 2, even though it's a breaking change. it just seems more efficient to me since the code is probably very similar. with things like this is when i which we could do `rotate -cw` and `rotate -ccw`. maybe we could make it `rotate --left/-l` or `rotate --right/-r`?

jntrnr:
👍 for one command

jntrnr:
Looks good!

PR #881, opened 2022-01-29
Variables should error on use rather than value span

jntrnr:
# Description

Before:

```
Error: nu::shell::type_mismatch (link)

  × Type mismatch during operation.
   ╭─[entry #1:1:1]
 1 │ benchmark { echo a | each { $it + 10 } | length }
   ·                  ┬              ┬ ─┬
   ·                  │              │  ╰── int
   ·                  │              ╰── type mismatch for operator
   ·                  ╰── string
   ╰────
```

Now:
```
Error: nu::shell::type_mismatch (link)

  × Type mismatch during operation.
   ╭─[entry #8:1:1]
 1 │ benchmark { echo a | each { $it + 10 } | length }
   ·                             ─┬─ ┬ ─┬
   ·                              │  │  ╰── int
   ·                              │  ╰── type mismatch for operator
   ·                              ╰── string
   ╰────
```

Fwiw: if we can in the future, I'd like to have both, but for now the use site is closer to the error and easier to debug.

# Tests

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -301,10 +301,7 @@ pub fn eval_expression(
             let block = engine_state.get_block(*block_id);
 
             for var_id in &block.captures {
-                captures.insert(
-                    *var_id,
-                    stack.get_var(*var_id)?, //.map_err(|_| ShellError::VariableNotFoundAtRuntime(expr.span))?,
-                );
+                captures.insert(*var_id, stack.get_var(*var_id, expr.span)?);
             }
             Ok(Value::Block {
                 val: *block_id,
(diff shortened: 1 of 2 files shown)


kubouch:
I think having spans pointing at the origin is useful as well so you can pinpoint the root cause. The problem is if there is _only_ the origin span.

We might want to standardize the error spans a little bit into, e.g., the error span and the origin span by keeping some naming convention for example.

jntrnr:
@kubouch - yeah, I was thinking if we shrunk Spans into SpanIds (which are a lookup into a list of Spans), then we could fit both the origin and use span in the same space as the current Span.

We'll need to play with it, but I agree that we actually want both.

jntrnr:
Actually... I might have an idea. Moving this to draft.

jntrnr:
Darn. I thought we might be able to mix in the expression span with the value span when we create errors, but this turns out to either be lots of targeted fixes or lots of breakage in how we write commands to adjust to the new style. Let's go with this fix for now, but keep thinking about using both spans in the future.

Issue #869, opened 2022-01-28
helper for datetime conversions


fdncred:
Something like this would be helpful as a `into datetime --list` type functionality. I know i can never remember these things.
https://www.strfti.me/
https://strftime.org/

voidash:
Hello, this is going to be my first contribution to this project. When i peeked into the code, i found out that the implementation is using `std::chrono` so  i built a cheatsheet from [https://docs.rs/chrono/0.3.1/chrono/format/strftime/index.html](https://docs.rs/chrono/0.3.1/chrono/format/strftime/index.html) 
Here is the help message for list

```
 -l, --list
      lists strftime cheatsheet
```
I used `Value::Record` to show the output. Here is the sample output
```
/home/cdjk/ >  '2020-09-04_16:39:18+0000' | into datetime --list                                       
╭──────┬─────────────────────────────────────────────────────────────────────────────────────────╮
│ code │ Description                                                                             │
│ %F   │ Year-month-day format. Same to %Y                                                       │
│ %Y   │ The full proleptic Gregorian year, zero-padded to 4 digits(Example: 2001).              │
│ %C   │ The proleptic Gregorian year divided by 100, zero-padded to 2 digits. (Example: 20)     │
│ %y   │ The proleptic Gregorian year modulo 100, zero-padded to 2 digits. (Example: 01)         │
│ %m   │ Month number (01--12), zero-padded to 2 digits.(Example: 07)                            │
│ %b   │ Abbreviated month name. Always 3 letters.(Example: Jul)                                 │
│ %B   │ Full month name. Also accepts corresponding abbreviation in parsing.(Example: July)     │
│ %h   │ Same to %b                                                                              │
│ %d   │ Day number (01--31), zero-padded to 2 digits.(Example: 08)                              │
│ %e   │ Same to %d                                                                              │
│ %a   │ Abbreviated weekday name. Always 3 letters.(Example: Sun)                               │
│ %A   │ Full weekday name. Also accepts corresponding abbreviation in parsing.(Example: Sunday) │
│ %w   │ Sunday = 0, Monday = 1, ..., Saturday = 6.(Example: 0)                                  │
│ %u   │ Monday = 1, Tuesday = 2, ..., Sunday = 7. (ISO 8601)(Example: 7)                        │
```

fdncred:
@voidash yes, good work. i love this. i left some comments on the pr. thank you very much!!!!

fdncred:
closed due to #883 yay!

[...]

PR #883, opened 2022-01-29
Add strftime cheatsheet for `into datetime` (#869)

voidash:
# Description
```

 > into datetime --list                                       
╭──────┬─────────────────────────────────────────────────────────────────────────────────────────╮
│ code │ Description                                                                             │
│ %F   │ Year-month-day format. Same to %Y                                                       │
│ %Y   │ The full proleptic Gregorian year, zero-padded to 4 digits(Example: 2001).              │
│ %C   │ The proleptic Gregorian year divided by 100, zero-padded to 2 digits. (Example: 20)     │
│ %y   │ The proleptic Gregorian year modulo 100, zero-padded to 2 digits. (Example: 01)         │
│ %m   │ Month number (01--12), zero-padded to 2 digits.(Example: 07)                            │
│ %b   │ Abbreviated month name. Always 3 letters.(Example: Jul)                                 │
│ %B   │ Full month name. Also accepts corresponding abbreviation in parsing.(Example: July)     │
│ %h   │ Same to %b                                                                              │
│ %d   │ Day number (01--31), zero-padded to 2 digits.(Example: 08)                              │
│ %e   │ Same to %d                                                                              │
│ %a   │ Abbreviated weekday name. Always 3 letters.(Example: Sun)                               │
│ %A   │ Full weekday name. Also accepts corresponding abbreviation in parsing.(Example: Sunday) │
│ %w   │ Sunday = 0, Monday = 1, ..., Saturday = 6.(Example: 0)                                  │
│ %u   │ Monday = 1, Tuesday = 2, ..., Sunday = 7. (ISO 8601)(Example: 7)                        │
```
  
# Tests

### DIFF
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -197,240 +197,916 @@ fn operate(
 }
 
 fn generate_strfttime_list(head: Span) -> Value {
-    let mut code = vec![];
-    let mut description = vec![];
-    code.push("code".into());
-    description.push(Value::String {
-        val: "Description".into(),
-        span: head,
-    });
-    code.push("%F".into());
-    description.push(Value::String {
-        val: "Year-month-day format. Same to %Y".into(),
-        span: head,
-    });
-    code.push("%Y".into());
-    description.push(Value::String {
-        val: "The full proleptic Gregorian year, zero-padded to 4 digits(Example: 2001).".into(),
-        span: head,
-    });
-    code.push("%C".into());
-    description.push(Value::String {
-        val: "The proleptic Gregorian year divided by 100, zero-padded to 2 digits. (Example: 20)"
-            .into(),
-        span: head,
-    });
-    code.push("%y".into());
-    description.push(Value::String {
-        val: "The proleptic Gregorian year modulo 100, zero-padded to 2 digits. (Example: 01)"
-            .into(),
-        span: head,
-    });
-    code.push("%m".into());
-    description.push(Value::String {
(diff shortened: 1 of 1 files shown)


fdncred:
nice! i just added #869 to remember this yesterday. 

one suggestion, if you want it to be a table (which is just a list of records), you can try and make the return a Value::List of Records instead of adding "code and description" as kind of a header above. so, i think you could do something like:

```
Value::List {
  vals: vec![Value::Record {
        cols: code,
        vals: description,
        span: head,
    }],
  span: head
}
```
or perhaps like the `bool` example https://github.com/nushell/engine-q/blob/1a259706451993c3ce811d2b25315260c8b53638/crates/nu-command/src/conversions/into/bool.rs#L46-L76

My code may not be precise but you should be able to fish around in the code-base to find a list of records.

voidash:
Thanks for pointing it out; I actually wanted a table. I looked into `into/bool.rs` and figured it out. Here is the sample output 
```
╭────┬───────────────┬────────────┬─────────────────────────────────────╮
│ #  │ Specification │  Example   │             Description             │
├────┼───────────────┼────────────┼─────────────────────────────────────┤
│  0 │ %Y            │ 2001       │ The full proleptic Gregorian year,  │
│    │               │            │ zero-padded to 4 digits             │
```

fdncred:
That output looks great! It looks so tedious writing all those `Value::Record` statements. It says to me that we need a helper function because that really looks like a pain to write. haha.

I'll land this as soon as the CI goes green. Thanks for all your hard work!!

fdncred:
i love this so much! ❤️

PR #890, opened 2022-01-30
add help --find to help doc

stormasm:
This lets everyone know when they type the help command that

```rust
help --find <text to search>
```

is a very valuable way to search through all of help.
I can never remember this myself so I thought other users would
want to know about it too.


onthebridgetonowhere:
That's pretty useful indeed, thanks for adding this!

PR #891, opened 2022-01-30
Port update cells command

rusty-jules:
# Description

Hello! I have done a port of `update cells`, even though I did not sign up for it. This was really just a learning experience for me to dip my toe into contributing to nushell (I love this tool!!), but only after writing this did I notice that this port had been assigned to @kubouch (I'm assuming) in nushell/nushell#4356, although it had not been assigned in nushell/engine-q#242 which is what I had been looking at. Did not mean to not follow protocol so I don't mind closing if this is unhelpful.
  
# Tests

- [ ] `cargo clippy --all --all-features -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style <-- **this failed [here](https://github.com/nushell/engine-q/blob/main/crates/nu-system/src/macos.rs#L329)**

## Notes

I really went into this blind and probably did some things in a non-idiomatic way, but the tests from the original are passing. There are currently two `FIXME`s [[one](https://github.com/nushell/engine-q/compare/main...rusty-jules:port-update-cells?expand=1#diff-a0f27b3988e2858819fe619d4855f48e1cb672dae2d5c20feafb153737dd4349R205), [two](https://github.com/nushell/engine-q/compare/main...rusty-jules:port-update-cells?expand=1#diff-a0f27b3988e2858819fe619d4855f48e1cb672dae2d5c20feafb153737dd4349R247)] where I had questions.


kubouch:
Haha, that's fine, I haven't started yet! Thanks for the effort!

rusty-jules:
Whoops, a last minute attempt to make the code prettier broke it. I've reverted it back to a less-pretty `if` statementj. I promise it passes this time! So much for a good first impression 😅

jntrnr:
@rusty-jules - heh, no worries. For that clippy error, you'll need to update your code following clippy's suggestions. Instead of unwrapping, we have a standard error we use everywhere: ShellError. You can check for the possible error and return ShellError, or, if the error is impossible, use `.expect`. The other clippy warning I think you can turn your `.into_iter()` to just `.iter()`

crates/nu-command/src/filters/update_cells.rs
```diff
@@ -0,0 +1,265 @@
...
+    stack: &mut Stack,
+    block: &Block,
+    span: Span,
+) -> Value {
+    if let Some(var) = block.signature.get_positional(0) {
+        if let Some(var_id) = &var.var_id {
+            stack.add_var(*var_id, val.clone());
+        }
+    } else {
+        // FIXME: Do we set the $it variable here?
```
kubouch, line 247: I think it's OK to leave this empty.

crates/nu-command/src/filters/update_cells.rs
```diff
@@ -0,0 +1,265 @@
...
+                                    &mut self.stack,
+                                    &self.block,
+                                    span,
+                                ),
+                            })
+                            .collect(),
+                        cols,
+                        span,
+                    }),
+                    // FIXME: is stepping into nested lists desired?
```
kubouch, line 205: I wouldn't step into nested lists. 

If the input is a list, each of its cells should be processed. If some of the cells are lists, they should be still treated as a single cell.

crates/nu-command/src/filters/update_cells.rs
rusty-jules: Great, would this still be a todo or should we remove it completely?

kubouch:
@rusty-jules I left a comment. Otherwise looks good to me.

kubouch:
Looks good thanks a lot!

PR #894, opened 2022-01-30
Allow viewing the source code of blocks

kubouch:
# Description

## Features

This PR adds an experimental `view` command that lets you view the source code of:
* a block 
* a module
* a custom command

## Caveats

* Currently, viewing individual blocks of `export env` is not possible -- you need to view the whole module. However, viewing `export def` and `export def-env` is possible (e.g., using `view "module-name def-name").
* It shows only the block part. E.g., the `def name [args...]` parts will be lost.
  
# Tests

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -330,7 +330,7 @@ pub fn create_default_context(cwd: impl AsRef<Path>) -> EngineState {
 
         // Experimental
         bind_command! {
-            View,
+            ViewSource,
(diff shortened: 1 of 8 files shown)


vwkd:
“View” seems very generic choice for internal source code and might conflict in the future with some actual viewing functionality. Maybe a more explicit name like “View-source” would be better?

jntrnr:
Ahh, I like `view-source`, too. Reminds me of viewing source on a webpage :grinning:

PR #898, opened 2022-01-31
Use 'table' during internal->external

jntrnr:
# Description

This switches from outputting debug text to translating structured data via `table` (without color) when sending between an internal command and an external command.

Things this allows you to do:

```
> ls | less
```
```
> ls | grep darren
│  6 │ darren.nu        │ file │     154 B │ 3 weeks ago  │
│ 13 │ source_darren.nu │ file │      50 B │ 2 weeks ago  │
```
And so on.

In short, it helps play a little nicer at the boundary if people just want to fall back onto their current toolset with what they're looking at on the screen.
  
# Tests


fdncred:
I initially was going to object to this because it removes the color. However, since one can still do `ls | table | less` and it outputs color, I like it. :)

I think it's clever how, instead of stripping color, you just faked the config setting, using what we already have built in. nice.

PR #910, opened 2022-02-02
Only grab the signature once in eval_call

sophiajt:
# Description

This runs one step to get the signature at the beginning of `eval_call`.

Benchmarks like:

```
> def foo [] { 10 }; benchmark { 1..100000 | each { $it + (foo) } | length }
```

Are roughly 10% faster now.
  
# Tests


jntrnr:
What in the world. This also makes gradient slower now(??!) even though in theory it should be noticeably faster also.

Will switch to draft so we don't land before we can dig in further.

jntrnr:
Closing - need to measure before I optimise.

Issue #907, opened 2022-02-01
let-env works inside a def-env block, but hide does not


ajeetdsouza:
One would assume that the two commands have the same semantics:

```console
$ def-env set-foo [] { let-env foo = "foo" }
$ set-foo
$ env # foo is visible
$ def-env unset-foo [] { hide foo }
$ unset-foo
$ env # foo is still visible
```

edit: follow the [discord conversation here](https://discord.com/channels/601130461678272522/759901889978040331/938064315267829800)

PR #921, opened 2022-02-03
Allow def-env to hide environment variables

kubouch:
# Description

Fixes https://github.com/nushell/engine-q/issues/907

```
> let-env FOO = "foo"

> def-env hide-foo [] {
    hide FOO
}

> $env.FOO
# Error, FOO not found.
```
  
# Tests

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -133,14 +133,28 @@ fn eval_call(
                 }
             }
         }
+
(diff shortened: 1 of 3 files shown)


kubouch:
Please don't merge, I realized I can just insert callee's stack.env_vars to the caller's stack and not do all this shenanigans. 

(I'm not sure how to convert to a draft PR on the phone.)

kubouch:
Ah, nevermind, it was actually correct :-D

PR #932, opened 2022-02-04
fix into filesize tests and filesize

fdncred:
# Description

fix `into filesize` and it's tests
  
# Tests

### DIFF
--- a/crates/nu-command/src/conversions/into/filesize.rs
+++ b/crates/nu-command/src/conversions/into/filesize.rs
@@ -110,28 +110,36 @@ fn into_filesize(
     )
 }
 
-pub fn action(input: &Value, span: Span) -> Value {
+pub fn action(input: &Value, _span: Span) -> Value {
+    let value_span = input.span().expect("unable to get input value span");
     match input {
         Value::Filesize { .. } => input.clone(),
-        Value::Int { val, .. } => Value::Filesize { val: *val, span },
+        Value::Int { val, .. } => Value::Filesize {
+            val: *val,
+            span: value_span,
+        },
         Value::Float { val, .. } => Value::Filesize {
             val: *val as i64,
-            span,
+            span: value_span,
         },
-        Value::String { val, .. } => match int_from_string(val, span) {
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/conversions/into/filesize.rs
```diff
@@ -110,28 +110,36 @@ fn into_filesize(
     )
 }
 
-pub fn action(input: &Value, span: Span) -> Value {
+pub fn action(input: &Value, _span: Span) -> Value {
+    let value_span = input.span().expect("unable to get input value span");
```
jntrnr, line 114: this will cause nu to crash if anyone passes an error value in. I think instead, you want to just call `if let Some(span) = input.span() { ... }`

fdncred: ok. how's it look now?

crates/nu-command/src/conversions/into/filesize.rs
```diff
@@ -110,28 +108,43 @@ fn into_filesize(
...
+                ),
+            },
+        }
+    } else {
+        Value::Error {
             error: ShellError::UnsupportedInput(
-                "'into filesize' for unsupported type".into(),
-                span,
+                "unable to find span on 'into filesize' value".into(),
+                Span::test_data(),
```
jntrnr, line 141: I'd use the span you're getting into the function up above, `span`. The `Span::test_data()` command should only be used when writing test data, but here we're trying to show a real error to the user.

jntrnr: I should not you removed that bit, but it'd be good to bring it back and use it here

crates/nu-command/src/conversions/into/filesize.rs
jntrnr: definitely closer. I think one last tweak and you've got it

crates/nu-command/src/conversions/into/filesize.rs
fdncred: ok, i re-undid-removed-put-back span again ;)

PR #946, opened 2022-02-05
fix test math/avg.rs can_average_bytes

stormasm:
This fixes the test **can_average_bytes**

in the file math/avg.rs

### DIFF
--- a/crates/nu-command/tests/commands/math/avg.rs
+++ b/crates/nu-command/tests/commands/math/avg.rs
@@ -14,14 +14,12 @@ fn can_average_numbers() {
     assert_eq!(actual.out, "101.5")
 }
 
-// FIXME: jt: needs more work
-#[ignore]
 #[test]
 fn can_average_bytes() {
     let actual = nu!(
         cwd: "tests/fixtures/formats",
-        "ls | sort-by name | skip 1 | first 2 | get size | math avg | format \"{$it}\" "
+        "ls | sort-by name | skip 1 | first 2 | get size | math avg | to json -r"
     );
 
-    assert_eq!(actual.out, "1.6 KB");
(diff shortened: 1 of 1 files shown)


fdncred:
I looked at this test yesterday and saw that it was a deficiency in the `format` command which led me to conclude the `format` command needed to be fixed. I'm not suggesting that anyone did anything wrong but I bring this up to demonstrate the vagueness of 'fixing' tests. There are a lot of tests that can be fixed by changing what they do but then we have missing functionality.

another test that fits this description is the base64 decode that has a single quote in the string. base64 will decoded it just fine but our test suite won't handle it yet. so should we fix the test to decode some other data that doesn't have a single quote in the middle of it or should we fix the test harness?

stormasm:
@fdncred yes fully agreed with your comment...

that is why I decided not to land the test myself but rather let you or JT or whomever make the call on how to move forward....

I looked at a bunch of other tests yesterday on my "testing rampage" and found a lot of the tests could be fixed but I noticed especially with most of the remaining math commands that I was not sure how we wanted to go forward.

Fix functionality in concert with fixing the tests ?  So I decided to hold off on fixing any more tests for now, as most of the remaining ones I looked at and was not sure how to proceed.

fdncred:
ya, @jntrnr and I talked a little about this yesterday. I'm not sure we reached any consensus but the two options are.

1. if we want to maintain backwards compatibility we must fix the command
2. if we don't care about backwards compatibility we can change the test so it passes.
3. there are some tests that probably don't fall into either of these and are other options, mostly of which are add the command since it doesn't exist or fix the command because it's obvious that we need that fix.

my vote is for number 1, but I figure that's the more difficult route unfortunately.

